Issue #33, opened 2015-07-26
`grapql` string context macro to create parsed document and verify query at compile time


OlegIlyenko:

dlreeves:
Not sure where this lands on your priorities, but I have been messing with this over the last couple days. It's been a while since I've done any Scala code, but I was able to get something that is able to parse a graphql document that is a literal string at compile time.

If you are taking pull request I'll try to clean it up and send one your way

OlegIlyenko:
Of course, I would appreciate this!  Even if it's not perfect yet, we can improve over the time, but we already will have something functional to begin with.

PR #56, opened 2015-08-23
#33 - Macro for parsing graphql documents at compile time

dlreeves:
This addresses issue #33 by adding a LiteralGraphQLStringContext  that adds a graphql method. This allows you to use string interpolation to parse a string literal into a sangria.ast.Document. This is done via a simple macro that calls QueryParse.parse on the string literal pulled from the StringContext call.

The majority of the code is in implementing the Liftable type class for all the ASTNodes. This requires quite a bit of boilerplate, but should be fine for now. Perhaps there is something available that implements this type class automatically for case classes?

This macro allows us to parse the string literal at compile time. Compilation will fail if the string cannot be parsed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,6 +22,9 @@ libraryDependencies ++= Seq(
   "org.scalatest" %% "scalatest" % "2.2.4" % "test"
 )
 
+// macros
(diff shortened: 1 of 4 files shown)


OlegIlyenko:
Looks great, thanks a lot! Sorry for this failing coveralls check, I haven't configured it properly yet.

PR #116, opened 2016-02-29
Travis containers and caching

jonas:
I didn't try out `before_cache` but have been using a similar setup to enable Travis' caching in a different project to speed up builds. Split the indentation change in a separate commit in case that matters I can either revert or squash.

Also adds a small improvement to remove some boilerplate from `QueryRendererSpec`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: scala
 scala:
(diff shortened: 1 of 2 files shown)


OlegIlyenko:
Nice improvements, thanks a lot!

PR #135, opened 2016-05-14
add support for case objects defined in companion object

joprice:
This is not the prettiest implementation, but I don't know the macro apis well enough to know the best way to get the enclosing object, if any. For instance, this wouldn't work for arbitrary levels of nesting. But, I think this single level is a common use case (I have it in my own codebase and would have to rewrite other code in order to use this lib).


src/test/scala/sangria/macros/derive/DeriveEnumTypeMacroSpec.scala
```diff
@@ -31,6 +39,11 @@ class DeriveEnumTypeMacroSpec extends WordSpec with Matchers {
   case object MegaOrangeAnnotated extends FruitAnnotated
 
   "Enum derivation" should {
+    "support nested enum instances"  in {
+      val enum = deriveEnumType[Difficulty]()
+      enum.name should be ("Difficulty")
```
OlegIlyenko: It would be great if you could add assertions for all 3 enum values as well (that there are only 3 available their names are correct). I often had this issue where `knownDirectSubclasses` does not find all of the subtypes, so I normally always try to verify values as well.

OlegIlyenko:
Thanks a lot for putting this together!

> this wouldn't work for arbitrary levels of nesting. But, I think this single level is a common use case

I agree, I think it's the most common way to define the enum values and one level of nesting should suffice.

PR #138, opened 2016-05-23
Added new materialization logic method to allow abstract type resolution

jlawrienyt:
This PR adds a new `classForType` method to the `MaterializationLogic` trait which will allow proper resolution of any Union or Interface type appearing in a materialized schema.  Implementations could simply be calling the `get` method on a `Map[String, Class[_]]` instance or could be a pattern match like below:
```
def classForType(name: String): Option[Class[_]] => name match {
  case "Type1" => Some(classOf[Type1])
  case "Type2" => Some(classOf[Type2])
  case _           => None
}
```

### DIFF
--- a/src/main/scala/sangria/schema/IntrospectionSchemaMaterializer.scala
+++ b/src/main/scala/sangria/schema/IntrospectionSchemaMaterializer.scala
@@ -9,6 +9,7 @@ import sangria.renderer.SchemaRenderer
 import sangria.validation.Violation
 
 import scala.collection.concurrent.TrieMap
+import scala.reflect.ClassTag
 import scala.util.{Try, Failure, Success}
 
 class IntrospectionSchemaMaterializer[Ctx, T : InputUnmarshaller] private (introspectionResult: T, logic: MaterializationLogic[Ctx]) {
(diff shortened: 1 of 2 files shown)


jlawrienyt:
Oleg, this looks great.  In my experiments, I was trying to just use the Play JSON AST instead of concrete types, so your branch will definitely help with that.  Also, I'd definitely prefer to materialize the schema based on the IDL syntax instead of the JSON since it is much less verbose.  Thanks!

I am curious though,  any reason not to also allow an `isInstanceOf` override in one of the ObjectType factory methods, if creating the schema in code?

OlegIlyenko:
Great, than I will proceed with my plan and merge the unified builder approach.

> I am curious though, any reason not to also allow an isInstanceOf override in one of the ObjectType factory methods, if creating the schema in code?

I guess it was an oversight. In a long term it should be possible, of course. I will look how I can enable it later.

OlegIlyenko:
Just added a test that demonstrates how this new schema builder works:

https://github.com/sangria-graphql/sangria/blob/ast-materializer/src/test/scala/sangria/schema/AstSchemaMaterializerSpec.scala#L774

I overridden the `objectTypeInstanceCheck` and distinguish objects based on the `type` key in the map.

OlegIlyenko:
OK I will close this PR in favor of new changes (I already merged everything into master). You work was very helpful, thanks again!

PR #173, opened 2016-10-24
Add coercion checks for more boundary values

jonas:
Check coercion of Double.{NaN,PositiveInfinity,NegativeInfinity} as well
as empty string.


jonas:
Maybe these are a bit pointless (didn't look at the implementation) but at least it spells out expected values and mirrors the changes done to the `coerceOutput` tests.

OlegIlyenko:
Nice, thanks a lot! Indeed it may not look super useful now, but I think that in a long term it may be helpful, especially if this behavior of input types will change for some reason

Issue #186, opened 2016-11-21
Invalid operation name should be considered a client-side error, thus should be a `QueryAnalysisError`


OlegIlyenko:
In this code it is considered a generic execution error: https://github.com/sangria-graphql/sangria/blob/90c7869738a19be04fa770358283da33afe444e7/src/main/scala/sangria/execution/Executor.scala#L118-L129

We probably can introduce an specific exception class that extends `QueryAnalysisError` for it.

PR #187, opened 2016-11-21
186: add custom subclass for unknown operation name

mattfenwick:
Based on #186 .

This is my first PR, so I have a bunch of questions:

 - [x] does the code follow established conventions (indentation, naming, etc.)?

 - [x] did I extend the right class and mix-in the right trait?

 - [x] should a source mapper and position be included, as in `OperationSelectionError`?

 - [x] I'm not sure where and how to add tests for the new error

Happy to fix anything, no matter how minor -- thanks for taking a look! :)

### DIFF
--- a/src/main/scala/sangria/execution/ExecutionError.scala
+++ b/src/main/scala/sangria/execution/ExecutionError.scala
@@ -44,3 +44,6 @@ case class QueryReducingError(cause: Throwable, exceptionHandler: Executor.Excep
 
 case class OperationSelectionError(message: String, eh: Executor.ExceptionHandler, sm: Option[SourceMapper] = None, pos: List[Position] = Nil)
   extends ExecutionError(message, eh, sm, pos) with QueryAnalysisError
+
+case class UnknownOperationNameError(opName: String, eh: Executor.ExceptionHandler) extends ExecutionError(
+  s"Unknown operation name $opName", eh) with QueryAnalysisError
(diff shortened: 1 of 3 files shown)


OlegIlyenko:
Now that I looks at it again, I noticed [`OperationSelectionError`](https://github.com/sangria-graphql/sangria/pull/187/files#diff-dcf69c220271eed098e7e6d5e9333a1dR114). Maybe we can adopt it for this use-case as well, it sounds pretty suitable. Maybe we can add addition parameter for it `operationName: Option[String]` (in case somebody will use it problematically) and provide a different message? `s"Unknown operation name $opName"` sounds great, but I would suggest to surround dynamic part with `'` (single quote) so that it's easier for user to distinguish it.

What do you think?

> should a source mapper and position be included, as in OperationSelectionError?

Even though we have source mapper in scope (so we know where this query comes from), we don't have a position, so I would skip both.

> does the code follow established conventions (indentation, naming, etc.)?

Everything looks wonderful! :)

> I'm not sure where and how to add tests for the new error

I would suggest to add some test close to this one:

https://github.com/sangria-graphql/sangria/blob/25ce3202a456687c50c36d99f0c09739063b1419/src/test/scala/sangria/execution/ExecutorSpec.scala#L408-L416

OlegIlyenko:
BTW, thanks for working on this PR!

mattfenwick:
👍 thanks for the suggestions -- I agree with your approach, and I'll make those changes.

mattfenwick:
Okay, I swapped in `OperationSelectionError` and added a unit test.  Thanks again for the suggestions! 

> Maybe we can add addition parameter for it operationName: Option[String] (in case somebody will use it problematically) and provide a different message? 

I'm not sure if I did that right -- is the current patch what you had in mind?

OlegIlyenko:
Nice! Thanks a lot again!

PR #203, opened 2017-01-13
upgrade to sbt-scoverage 1.5.0

sullis:


OlegIlyenko:
Sweet! Was not aware that it is already released for 2.12

PR #215, opened 2017-02-15
Field name transformer for `deriveObjectType`

ostronom:
There is an `UppercaseValues` setting for `deriveEnumType`, but no means to transform field names in `deriveObjectType`. This PR is my attempt to provide such methods. I don't really know how this issue should really be approached and this is my first take on macros :)

### DIFF
--- a/src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
+++ b/src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
@@ -3,6 +3,7 @@ package sangria.macros.derive
 import sangria.execution.deferred.Deferred
 import sangria.macros._
 import sangria.schema.{Action, Context}
+import sangria.util.StringUtil
 
 import scala.concurrent.Future
 import scala.reflect.macros.blackbox
@@ -105,9 +106,14 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
             case (acc, _) ⇒ acc
           }
 
+          val objectTypeName: c.universe.Tree = config.collect{case MacroTransformFieldNames(fnt) => fnt}.lastOption match {
+            case Some(fnt) => q"$fnt(${configName orElse annotationName getOrElse q"$name"})"
+            case None => configName orElse annotationName getOrElse q"$name"
+          }
+
           q"""
             sangria.schema.Field[$ctxType, $valType, $actualFieldType, $actualFieldType](
-              ${configName orElse annotationName getOrElse q"$name"},
+              $objectTypeName,
               implicitly[sangria.macros.derive.GraphQLOutputTypeLookup[$actualFieldType]].graphqlType,
               ${configDescr orElse annotationDescr},
               $args,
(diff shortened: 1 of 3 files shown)


src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
```diff
@@ -105,9 +106,14 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
             case (acc, _) ⇒ acc
           }
 
+          val objectTypeName: c.universe.Tree = config.collect{case MacroTransformFieldNames(fnt) => fnt}.lastOption match {
```
OlegIlyenko: I guess it should be called `fieldName`?

src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
```diff
@@ -105,9 +106,14 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
...
+            case Some(fnt) => q"$fnt(${configName orElse annotationName getOrElse q"$name"})"
+            case None => configName orElse annotationName getOrElse q"$name"
```
OlegIlyenko: maybe we can extract `configName orElse annotationName getOrElse q"$name"` in a val, just to avoid the code duplication?

src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
```diff
@@ -357,4 +366,5 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
   case class MacroExcludeFields(fieldNames: Set[String], pos: Position) extends MacroDeriveObjectSetting
   case class MacroAddFields(fields: List[Tree]) extends MacroDeriveObjectSetting
   case class MacroReplaceField(fieldName: String, field: Tree, pos: Position) extends MacroDeriveObjectSetting
+  case class MacroTransformFieldNames(transformerName: Symbol) extends MacroDeriveObjectSetting
```
OlegIlyenko: Using a `Symbol` here concerns me a bit. Maybe it would be better to define it as `transformer: Tree`? WDYT?

src/test/scala/sangria/macros/derive/DeriveObjectTypeMacroSpec.scala
```diff
@@ -202,6 +202,19 @@ class DeriveObjectTypeMacroSpec extends WordSpec with Matchers with FutureResult
       tpe.fields(2).fieldType should be (IntType)
     }
 
+    "allow field names transformation" in {
+      val transformer1 = (s: String) => s.toUpperCase
+      val tpe = deriveObjectType[Unit, TestSubjectAnnotated](TransformFieldNames(transformer1))
+      tpe.fields.forall(f => f.name == transformer1(f.name)) shouldBe true
```
OlegIlyenko: If I understand correctly, `f.name` is already transform name. This assertion feels a bit unsafe for me. What do you think about explicitly asserting on every field? E.g.

```scala
tpe.fields.map(_.name) should (
  have(size(2)) and
  contain("ID") and 
  contain("MYLIST"))
```

For test assertions I prefer this more explicit style. It makes it easier for a reader to see the expected values.

OlegIlyenko: Or even simpler: 

```scala
tpe.fields.map(_.name).toSet should be (Set("ID", "MYLIST"))
```

ostronom:
I have two problems, which i want to solve with this:

1) Convert camelCase field names to snake_case ones
2) Drop prefixes from field names of classes generated by some tools (for example, i have a bunch of classes with field names starting with `Api` prefix, which is meaningless for the GraphQL-consumer)
3) I think that mapping function is more flexible way to introduce such conversions. For example, like this:

`case object SnakeCaseFieldNames extends TransformFieldNames((s: String) => s.toUpperCase)`

Thank you for deep review. I will update my PR according to it.

OlegIlyenko:
👍  Thanks for describing the use-case! Indeed, in this case we probably should go with generic `TransformFieldNames` instead of something more specific.

OlegIlyenko:
Since we already discussed it and due to inactivity, I will merge it now and do discussed improvements.

PR #216, opened 2017-02-17
Update README.md

helfer:
Was recently asked on slack if Sangria worked with Apollo. I thought it's best to clarify that directly in the Readme, in the same place where Relay is mentioned.


OlegIlyenko:
Great thanks a lot!

What do you think about making it a bit more generic?

```markdown
Sangria is a spec compliant GraphQL implementation, so it works out of the box with [Apollo](https://github.com/apollographql/apollo-client), [Relay](https://facebook.github.io/relay/), [GraphiQL](https://github.com/graphql/graphiql) and other tools and libraries.
```

Issue #218, opened 2017-02-25
Sangria does not have an OSGi compliant JAR


PhilAndrew:
Sangria Marshalling API and Sangria Streaming API and Sangria (this project) does NOT have an OSGi compliant JAR.

schrepfler:
What would need to happen so that it's OSGI compliant?

PhilAndrew:
Its not unusual to be non-OSGi compliant, its just that I'm trying to use it in an OSGi environment.

To show what the problem is firstly consider sangria-streaming-api.

Its META-INF/MANIFEST.MF file contains the following.

```
Manifest-Version: 1.0
Implementation-Title: sangria-streaming-api
Implementation-Version: 1.0.0
Specification-Vendor: org.sangria-graphql
Specification-Title: sangria-streaming-api
Implementation-Vendor-Id: org.sangria-graphql
Specification-Version: 1.0.0
Implementation-URL: http://sangria-graphql.org
Implementation-Vendor: org.sangria-graphql
```

That is not enough for OSGi, as we need to also include the imported and exported packages for the bundle to indicate to an OSGi container or tooling what packages the JAR wants to use and what packages it wants to make public for other people to use.

Consider camel-weather. Its MANIFEST.MF file is

```
Manifest-Version: 1.0
Export-Service: org.apache.camel.spi.ComponentResolver;component=weath
 er
Bundle-License: https://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: org.apache.camel.camel-weather
Archiver-Version: Plexus Archiver
Built-By: camel
Bnd-LastModified: 1485111615821
Specification-Title: Camel :: Weather
Implementation-Vendor-Id: org.apache.camel
Bundle-DocURL: https://www.apache.org/
Import-Package: org.apache.camel;version="[2.18,2.19)",org.apache.came
 l.impl;version="[2.18,2.19)",org.apache.camel.spi;version="[2.18,2.19
 )",org.apache.camel.util;version="[2.18,2.19)",org.apache.commons.htt
 pclient;version="[3.1,4.0)",org.apache.commons.httpclient.auth;versio
 n="[3.1,4.0)",org.apache.commons.httpclient.methods;version="[3.1,4.0
 )",org.slf4j;version="[1.7,2)",org.codehaus.jackson;version="[1.9,2)"
 ,org.codehaus.jackson.map;version="[1.9,2)",org.osgi.framework;versio
[...]

PR #222, opened 2017-03-01
Added SBT-OSGi to allow Sangria to generate an OSGi valid JAR artifact

PhilAndrew:
Added SBT-OSGi to allow Sangria to generate an OSGi valid JAR artifact on publish or publishLocal, this allowed Sangria to be used within an OSGi environment. Look at the file MANIFEST.MF inside the artifact JAR (result of publishLocal) before and after this change was added - to see the changes. The changes are that there are import and export packages defined there which allows OSGi to work out what imports and exports the JAR defines.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,6 +22,13 @@ scalacOptions ++= {
     Seq("-target:jvm-1.7")
 }
 
+lazy val sangriaOsgiSettings = osgiSettings ++ Seq(
+  OsgiKeys.exportPackage := Seq("sangria.*;version=${Bundle-Version}"),
+  OsgiKeys.privatePackage := Seq()
+)
+
+lazy val sangriaProject = project.in(file(".")).enablePlugins(SbtOsgi).settings(sangriaOsgiSettings:_*)
+
 libraryDependencies ++= Seq(
   // macros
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -22,6 +22,13 @@ scalacOptions ++= {
     Seq("-target:jvm-1.7")
 }
 
+lazy val sangriaOsgiSettings = osgiSettings ++ Seq(
+  OsgiKeys.exportPackage := Seq("sangria.*;version=${Bundle-Version}"),
+  OsgiKeys.privatePackage := Seq()
+)
+
+lazy val sangriaProject = project.in(file(".")).enablePlugins(SbtOsgi).settings(sangriaOsgiSettings:_*)
```
jonas: To follow the existing style this can be simplified to:

```sbt
enablePlugins(SbtOsgi)
OsgiKeys.exportPackage := Seq("sangria.*;version=${Bundle-Version}")
OsgiKeys.privatePackage := Seq()
```

OlegIlyenko:
I will close this PR for the moment being due to inactivity. I'm still pretty much open to adding an OSGi support, but it turns out to be a bit more involving. I would suggest keeping https://github.com/sangria-graphql/sangria/issues/218 open and continue the discussion there.

PR #228, opened 2017-03-11
rename TransformSpecial to VisitAnyField

searler:
Update to match pull request against macro-visit project


OlegIlyenko:
thanks! I will merge it as soon as I release the new version of macro-visit. I guess you can already use version `0.1.1` instead of snapshot.

PR #243, opened 2017-05-12
Create args builder

vishalkuo:
I'd like a helper be able to generate `Args` based on argument definitions and provided inputs. Generating Args manually can be rather tedious and a lot of the work seems like it can be abstracted away.

### DIFF
--- a/src/test/scala/sangria/schema/ArgsSpec.scala
+++ b/src/test/scala/sangria/schema/ArgsSpec.scala
@@ -1,0 +1,114 @@
+package sangria.schema
+
+import org.scalatest.{Matchers, WordSpec}
+
+import scala.collection.concurrent.TrieMap
+
+class ArgsSpec extends WordSpec with Matchers {
+
+
+  val NON_DEFAULT_ARGUMENT_NAME = "nonDefaultArgument"
+  val DEFAULT_ARGUMENT_NAME = "defaultArgument"
+  val OPTIONAL_ARGUMENT_NAME = "optionalArg"
+
+  val nonDefaultArgument = Argument(
+    name = NON_DEFAULT_ARGUMENT_NAME,
+    argumentType = IntType,
+    description = "Argument without default value"
+  )
+
+  val defaultArgument = Argument(
+    name = DEFAULT_ARGUMENT_NAME,
+    argumentType = OptionInputType(IntType),
+    defaultValue = 10,
+    description = "Argument with default value"
+  )
+
+  val optionalArgument = Argument(
+    name = OPTIONAL_ARGUMENT_NAME,
+    argumentType = OptionInputType(IntType),
+    description = "Optional argument"
+  )
+
+  "Args: Companion object" when {
+    "buildArgs" should {
+      "build with no arguments" in {
+        val args = Args.buildArgs(List.empty, Map.empty)
+        args.raw should be (Map.empty)
+        args.argsWithDefault should be (Set.empty)
+        args.optionalArgs should be (Set.empty)
+        args.undefinedArgs should be (Set.empty)
+        args.defaultInfo should be (TrieMap.empty)
+      }
+
+      "build with defined arguments" in {
+        val expectedMap = Map(NON_DEFAULT_ARGUMENT_NAME -> 9001)
+        val args = Args.buildArgs(List(nonDefaultArgument), expectedMap)
+        args.raw should be (expectedMap)
+        args.argsWithDefault should be (Set.empty)
+        args.optionalArgs should be (Set.empty)
+        args.undefinedArgs should be (Set.empty)
+        args.defaultInfo should be (TrieMap.empty)
+
+        args.arg(nonDefaultArgument) should be (9001)
+      }
+
+      "build with undefined arguments" in {
+        val args = Args.buildArgs(List(nonDefaultArgument), Map.empty)
+        args.raw should be (Map.empty)
+        args.argsWithDefault should be (Set.empty)
+        args.optionalArgs should be (Set.empty)
+        args.undefinedArgs should be (Set(NON_DEFAULT_ARGUMENT_NAME))
+        args.defaultInfo should be (TrieMap.empty)
+
+        an [NoSuchElementException] should be thrownBy args.arg(nonDefaultArgument)
+      }
+
+      "build with optional arguments and defined input" in {
+        val expectedMap = Map(OPTIONAL_ARGUMENT_NAME -> Some(9001))
+        val args = Args.buildArgs(List(optionalArgument), expectedMap)
+        args.raw should be (expectedMap)
+        args.argsWithDefault should be (Set.empty)
+        args.optionalArgs should be (Set(OPTIONAL_ARGUMENT_NAME))
+        args.undefinedArgs should be (Set.empty)
+        args.defaultInfo should be (TrieMap.empty)
+
+        args.arg(optionalArgument) should be (Some(9001))
+      }
+
+      "build with optional argument and undefined input" in {
+        val args = Args.buildArgs(List(optionalArgument), Map.empty)
+        args.raw should be (Map.empty)
+        args.argsWithDefault should be (Set.empty)
+        args.optionalArgs should be (Set(OPTIONAL_ARGUMENT_NAME))
+        args.undefinedArgs should be (Set.empty)
+        args.defaultInfo should be (TrieMap.empty)
+
+        args.arg(optionalArgument) should be (None)
+      }
+
+      "build with default values" in {
+        val args = Args.buildArgs(List(defaultArgument), Map.empty)
+        args.raw should be (Map.empty)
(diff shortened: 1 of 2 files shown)


OlegIlyenko:
Thanks a lot for the PR! 👍 

I like the idea. Though , I have some concerns about the implementation details. First is that it manually manipulates things like default values, undefined variables, etc. Default values in particular are very hard to deal with. in the current form it will not work as expected because default value can be not only a scala data structures, but also things like raw JSON values (and literally anything that has a `ToInput` type class defined for it). Also another concern is that some this code that handles arguments is duplicated and can be reused from other places.

I experimented with it a bit. I would suggest to implement this method in terms of `InputUnmarshaller` and reuse existing code to handle the actual argument coercion, default values, etc. Here is what I came up with:

https://gist.github.com/OlegIlyenko/3937e3b3f1902978fc73fc68b1f523ab

Advantage of this approach is that it is quite generic. So it's able to handle scenarios that you have described in the tests:

```scala
val a = Args(myArgs, "age" → 10, "size" → Map("width" → 10, "height" → 20))
```

but it also would be able to handle any input type that sangria supports. Here is an example of spray-json:

```scala
import sangria.marshalling.sprayJson._
import spray.json._

Args(myArgs,
  """{age": 10, "size": {"width": 10, "height": 20}}""".parseJson)
```

to make this approach possible, I just extracted `getArgumentValues` and made it more generic:

https://github.com/sangria-graphql/sangria/commit/d2bbf38a744f1bd9359b5b9d58b1a2560dfd7311

What do you think about this approach?

vishalkuo:
I appreciate the feedback and definitely agree with your approach. I will take a look at implementing this over the next couple of days. Thanks!

vishalkuo:
@OlegIlyenko I have a couple questions regarding this implementation and the `Args` behaviour itself.

Firstly, for arguments with a default value,  can they ever appear in the `undefinedArgs` Set in an Args instance? i.e. 

```scala
val args = Args(List(optionalArgument))
args.undefinedArgs should be (Set.empty) // or should this be Set("optionalArgumentName")
```
I ask because I would assume they are still defined by their default value. 

Secondly, should the defaultInfo set only be populated when no argument input is actually provided?

i.e.
```scala
val args = Args(List(defaultArgument), s"""{"$DEFAULT_ARGUMENT_NAME": 9001}""".parseJson)
args.defaultInfo should be (TrieMap.empty) // or should this be TrieMap(DEFAULT_ARGUMENT_NAME -> 10)
```

Finally, I'm having trouble building arguments using just a Map[String, Any] implementation, i.e. the following line: 

```scala
val args = Args(List(nonDefaultArgument), Map(NON_DEFAULT_ARGUMENT_NAME -> 9001))
```

yields
```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
[info]     at sangria.marshalling.MarshallingUtil$.convert(MarshallingUtil.scala:24)
[info]     at sangria.marshalling.MarshallingUtil$MarshaledConverter.convertMarshaled(MarshallingUtil.scala:37)
[info]     at sangria.schema.Args$.$anonfun$apply$2(Context.scala:351)
[info]     at sangria.schema.Args$$$Lambda$9914/1023504108.apply(Unknown Source)
[info]     at scala.Option.map(Option.scala:146)
[info]     at sangria.schema.Args$.$anonfun$apply$1(Context.scala:351)
[info]     at sangria.schema.Args$$$Lambda$9903/737526958.apply(Unknown Source)
[info]     at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
...
```

I'm wondering if I'm doing something wrong here... I've uploaded my current state of the code and [the current travis build](https://travis-ci.org/sangria-graphql/sangria/builds/232281423?utm_source=github_status&utm_medium=notification) should have more detailed stack traces

OlegIlyenko:
Thanks a lot for trying this approach!

> for arguments with a default value, can they ever appear in the undefinedArgs Set in an Args instance?

Yes, if user haven't specified them, then they still will be undefined from a query perspective. The default value will still be used correctly, but this small distinction might be important in some scenarios. "Undefined " generally means that user haven't specified it in the query in this context.

> Secondly, should the defaultInfo set only be populated when no argument input is actually provided?

If I remember correctly, `defaultInfo` is used in more complex scenarios where variables are used so it's a variable default and not field's default. 

> I'm having trouble building arguments using just a Map[String, Any] implementation

Thanks for noticing it. I looks like a bug. at least here should be a check: https://github.com/sangria-graphql/sangria-marshalling-api/blob/master/src/main/scala/sangria/marshalling/MarshallingUtil.scala#L23

I general, now I realised that `convertMarshaled` is a wrong approach here since it does not consider the type system. We need to write a specialized variation of it that considers argument type and does appropriate scalar coercion. I will give it a try.

OlegIlyenko:
@vishalkuo I updated the gist:

https://gist.github.com/OlegIlyenko/3937e3b3f1902978fc73fc68b1f523ab

It now reuses much better type-aware conversion from `ValueCoercionHelper`.

vkuo-coursera:
Thanks so much for the help, @OlegIlyenko I've updated the branch. Let me know what you think

vishalkuo:
Thanks so much for the help, @OlegIlyenko. I've updated the branch, let me know what you think

src/test/scala/sangria/schema/ArgsSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+import sangria.execution.AttributeCoercionError
+
+import scala.collection.concurrent.TrieMap
+import sangria.marshalling.sprayJson._
+import spray.json._
+
+class ArgsSpec extends WordSpec with Matchers {
+
+
+  val NON_DEFAULT_ARGUMENT_NAME = "nonDefaultArgument"
```
OlegIlyenko: I would suggest to use [standard scala naming convention for constants](http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods). For example `NonDefaultArgumentName` instead of `NON_DEFAULT_ARGUMENT_NAME `

OlegIlyenko:
@vishalkuo great, thanks a lot! I would suggest to add at least one more test with more complex input (for example an input object with nested input objects lists and optional fields with defaults).

WDYT?

vishalkuo:
Thanks for the feedback, @OlegIlyenko I've added some specs as you suggested. I built nested arguments as [follows](https://github.com/vishalkuo/sangria/blob/master/src/test/scala/sangria/schema/ArgsSpec.scala#L38-L43), let me know if you had something else in mind

src/test/scala/sangria/schema/ArgsSpec.scala
```diff
@@ -0,0 +1,248 @@
...
+             |  }
+             |}
+           """.stripMargin.parseJson
+
+        val args = Args(List(nestedParentArgument), json)
+        val fields = args.arg(nestedParentArgument).asJsObject.fields
+
+        fields(NonDefaultArgumentName) should be (JsNumber(1))
+        fields(DefaultArgumentName) should be (JsNumber(2))
+        an [NoSuchElementException] should be thrownBy fields(OptionalArgumentName)
```
OlegIlyenko: In this case I would suggest to assert it without throwing an exception. E.g.

```scala
fields.get(OptionalArgumentName) should be ('empty)
```

src/test/scala/sangria/schema/ArgsSpec.scala
```diff
@@ -0,0 +1,248 @@
...
+      "build without optional arguments" in {
+        val json =
+          s"""
+             |{
+             |  "$NestedParentArgumentName": {
+             |    "$NonDefaultArgumentName": 1,
+             |    "$DefaultArgumentName": 2
+             |  }
+             |}
+           """.stripMargin.parseJson
```
OlegIlyenko: For such simple json stricture (here and in other places above), maybe it would be better to use Json AST directly. E.g.

```scala
JsObject(NestedParentArgumentName → JsObject(
  NonDefaultArgumentName → JsNumber(1),
  DefaultArgumentName → JsNumber(2)))
```

WDYT?

src/test/scala/sangria/schema/ArgsSpec.scala
```diff
@@ -0,0 +1,248 @@
...
+
+        val args = Args(List(nestedParentArgument), json)
```
OlegIlyenko: In addition to json input, I would also suggest to add extra test with the `Map`-based input. For example, this code should produce the same result as JSON variation (and should be supported as well):

```scala
val args = Args(List(nestedParentArgument),
  NestedParentArgumentName → Map(
    NonDefaultArgumentName → 1, 
    DefaultArgumentName → 2))
```

vkuo-coursera:
@OlegIlyenko thanks for the review, I've updated my PR. Let me know what you think!

vishalkuo:
@OlegIlyenko thanks for the review! I've updated my PR. Let me know what you think

OlegIlyenko:
Wonderful 🎉 Thanks a lot for this PR!

Issue #245, opened 2017-05-16
QueryReducer depth analysis


Eluinhost:
Using a query like this:

```graphql
{
  complexity
  depth
  networkById(id:1) {
    servers {
      ip
    }
  }
}
```

shows the correct response as I was expecting back:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Max query depth 2 is reached. Query had a depth of 3"
    }
  ]
}
```

Slightly changing the query to this (moving complexity/depth to the end):

```graphql
{
  networkById(id:1) {
    servers {
      ip
    }
  }
  complexity
  depth
}
```

produces this response:

```json
{
  "data": {
    "networkById": null,
    "complexity": 5,
    "depth": 1
  }
}
```

I'm using this reducer if relevant

```scala
QueryReducer.measureDepth[SchemaContext] { (depth, ctx) ⇒
    if (depth > max) throw QueryTooDeepException(max, depth)

    ctx.copy(metadata = ctx.metadata.copy(depth = Some(depth)))
  }
```

I changed this line: https://github.com/sangria-graphql/sangria/blob/281910e58a6d26a80a555ff938be6dd7d6afc9b4/src/main/scala/sangria/execution/QueryReducer.scala#L125
to 
```scala 
Math.max(fieldAcc, childrenAcc)
``` 
and added 
```scala
calcDepth("""
       {
         nest {
           a
           b
           ab

           named(size: 1) {
             name

             ... on Cat {
               meows
             }

             ... on Dog {
               barks
             }
           }
         }

         nest {
           a
         }

         info
         a
       }
       """) should be (3)
```
in here: https://github.com/sangria-graphql/sangria/blob/281910e58a6d26a80a555ff938be6dd7d6afc9b4/src/test/scala/sangria/execution/QueryReducerSpec.scala#L609 and it *seemed* to be working as expected (or at least the test passed now) though I have no idea how this is supposed to work or if I'm just misunderstanding how depth is calculated/reduced

OlegIlyenko:
Great find! I think you are right. Would you mind making a PR with proposed change and a test(s)?

OlegIlyenko:
You are right. `reduceField` works like a fold, this means it also needs to consider the accumulator and merge it with the current value. `math.max` is the right call in this case. Thanks again!

PR #246, opened 2017-05-16
Fix `MeasureQueryDepth` reducer not keeping the largest depth found

Eluinhost:
From issue #245


OlegIlyenko:
This was fast! thanks a lot! 👍

PR #269, opened 2017-07-24
Add missing space in the description of the deprecation reason argument

jonas:
For reference, the space is not missing in the JS version: https://github.com/graphql/graphql-js/blob/831598ba76f015078ecb6c5c1fbaf133302f3f8e/src/type/directives.js#L169-L171


OlegIlyenko:
thanks!

PR #271, opened 2017-07-28
Built in scalars will now only be added to the schema if they are used

### LINKED ISSUE #270: Remove Unused Builtin Scalars From Schema
Currently, when performing an introspection query on a Sangria defined schema, all builtin scalars (standard GraphQL and custom Sangria instances) are returned regardless of whether they are actually used.  The ability to filter builtin when rendering a schema to SDL was recently added (issue #241), but that does impact clients that use introspection queries to fetch their schemas.

Potential solution: automatically filter out unused builtin scalars from the `Schema` instance

jlawrienyt:
This resolves #270


OlegIlyenko:
Beautiful! 🙌  Thanks a lot!

PR #274, opened 2017-08-04
Fixed directive definition rendering in query renderer

alexeygolev:
Currently if a directive doesn't have any arguments which is possible [per spec](https://facebook.github.io/graphql/#sec-Language.Directives) both compact and pretty printing doesn't have a space between the directive name and `on ...`
So it renders it like so for a directive named "custom"
```graphql
directive @customon FIELD
```
This PR fixes the behaviour so it renders
```graphql
directive @custom on FIELD
```
I didn't updated the kitchen sink test as it's breaking the expectations of the query parser test and I'll be honest I have no idea how to update that one


OlegIlyenko:
Nice, thanks a lot!

PR #275, opened 2017-08-08
[Minor] Add example test showing how the default action can be overridden

crypticmind:
Case class accessors will turn into `Value` actions, `Deferred`s into `DeferredValue`, and so on.

Here I'm just adding one more test to `DeriveObjectTypeMacroSpec` showing how the default action can be overridden by defining the corresponding implicit conversion for a type.

_yadda yadda_: I'm venturing in GraphQL using my own stubborn design, which turned the experience sort of into soap opera testing. I have an ADT `Ref[T]`, which can be `ID[T](id: String)` or `Persistent[T](id: String, obj: T)`. My repository will give me back `ID`s, but I wish to turn the `Value` actions for those `ID`s into `DeferredValue` actions which will resolve `Ref`s into `Persistent`s.

### DIFF
--- a/src/test/scala/sangria/macros/derive/DeriveObjectTypeMacroSpec.scala
+++ b/src/test/scala/sangria/macros/derive/DeriveObjectTypeMacroSpec.scala
@@ -11,6 +11,7 @@ import spray.json._
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
+import scala.language.implicitConversions
 import scala.util.{Success, Try}
 
 class DeriveObjectTypeMacroSpec extends WordSpec with Matchers with FutureResultSupport {
@@ -221,6 +222,24 @@ class DeriveObjectTypeMacroSpec extends WordSpec with Matchers with FutureResult
         have(size(2)) and
         contain("iD") and
         contain("mYlIsT"))
+    }
+
+    "allow overriding the action for a type" in {
+
+      case class A(id: String, b: B)
+      case class B(name: String)
+
+      implicit def BAction[Ctx](value: B): LeafAction[Ctx, B] = Value(value).map(b => b.copy(name = b.name.toUpperCase))
+      
+      implicit val BType = deriveObjectType[Unit, B]()
+      implicit val AType = deriveObjectType[Unit, A]()
+
+      val schema = Schema(AType)
+
+      val query = graphql"{ id, b { name } }"
+
+      Executor.execute(schema, query, root = A("1", B("foo"))).await should be (Map(
+        "data" → Map("id" → "1", "b" → Map("name" → "FOO"))))
     }
 
     "allow to set name, description, deprecationReason and fieldTags with annotations" in {


OlegIlyenko:
Thanks a lot for the suggestion! 

I'm not 100% about this approach. Even though scala implicit resolution makes it possible, `Action` is not intended to be used like this. I feel that it can be quite error-prone when used in this way.

In this partticulkar case I would normally suggest to use the `ReplaceField` macro setting:

```scala
implicit val BType = deriveObjectType[Unit, B](
  ReplaceField("name", Field("name", StringType, resolve = _.value.name.toUpperCase)))
```

It is more precise and explicit - you are replacing the field with the logic that is different from a default. I think that it better reflects the intention. What do you think?

crypticmind:
I see you point and I think it makes sense.

The reason I opted for the implicits it's because I'd like to generally alter how an Action is created for a type T, in order to change how that type T is supposed to be retrieved, and to avoid having to override fields of type T in each occurrence.

This is inline with what I mentioned in the PR comment, and you can see what I got so far [here](https://github.com/crypticmind/ssor/blob/master/framework/src/main/scala/crypticmind/ssor/model/package.scala).

Bridging an OO-design like graphql and a FP-oriented design isn't trivial, and my project is just a playground to test one way of accomplish that, which may or may not succeed. Which is way I'm open to the debate.

When you mentioned that it can be quite error-prone, are you discouraging the implicit resolution because of user usage concerns or do you have anything else in mind?

OlegIlyenko:
I see. I think my main concern is that it makes API definition more non-obvious. When it comes to API definition, I always prefer more explicit style. In my experience, one rarely has such a direct mapping from internal domain model and API model. For example, reference and ID fields normally have prefixes like `categoryRef` or `categoryId`. Ideally, you not only need to change the `resolve` logic but also rename exposed field to `category`. In some cases, you need to avoid the reference loading because of performance/design considerations and just expose an ID.

I am also very cautious when a simple import has such dramatic change on the exposed API. Just imagine that new team member looks at the schema file, reorganizes the imports and decides that import that brings `BAction` in scope is unused, so it is removed. The application compiles just fine, but suddenly the whole API now returns `Int` and `String` IDs instead of actual model objects :)

Having explicit `ReplaceField`s might produce a bit more boilerplate code, but I would argue that it serves better for the reader of the code since the reader can immediately see all fields that expand references and/or differ from the internal domain model. It also provides much more flexibility.

If you are using this approach for something very specific and well-defined within the application (like `Ref[T]`) and have an additional safety net around it, then it might be a reasonable solution. I definitely would not recommend this approach for more generic types, like `Int` or `String`. In the example from the test code, I would also prefer to uppercase `B.name` in the place where `BType` is defined, not on the side as an implicit conversion of a `LeafAction`. For me it also comes back to a separation between business logic of the application and infrastructure code. `Action`-based implicit conversions are just a syntax sugar to help with `resolve` function definitions, then are not intended as a mechanism for business logic definition.

Not to mention that in future `deriveObjectType` might avoid implicit resolution for `Action`s and generate them explicitly (there might be advantages to it). It will break this approach. So I'm quite uncomfortable defining it as an official feature. If it is a valuable featue/approach, then maybe we can model it in some different way?

Funny thing is that in our application we also have something very similar:  `Reference` type which has an `id` and `typeId`. But even if I wanted to, I can't really use this approach because of the mentioned reasons. But this is my personal experience. If it is possible, it would be interesting if you could validate this approach in real application. I wonder whether it still would be proven useful in your scenario.

It's a lot of text :) Wonder what you think about this line of reasoning?

crypticmind:
I believe I'm leaning towards a DDD style, in where I'm trying to derive an API from a model with as less intervention as possible, and you're more oriented towards explicit API design with explicit connection to the application services. A style that reminds me, for instance, of swagger API docs.

I don't have any fanaticisms whatsoever, on any side :)

Your explicit style seems more appropriate for the sangria project itself. With this, what I'm trying to say is that I don't think this PR applies anymore. This conversation was productive to me nevertheless to acquire a different point of view. 

For the time being, I will continue with my DDD style just to see how far it goes.

Thank you for your time!

OlegIlyenko:
👍 I will also keep this use-case in mind. Even though customization via implicit `Action` int not something I would like to highlight as a standard way to achieve this, I think that the use-case by itself is a valid one. So maybe we can think about good API for it? I will close this PR for the moment being, but please feel free to reopen it if you would like to continue the discussion (or maybe create a separate PR).

PR #276, opened 2017-08-10
Fix typo in error message.

jbgi:
This typo was vicious as it changed the original intended meaning. (just lost an hour because of it :smile: )


OlegIlyenko:
Thanks a lot!

PR #278, opened 2017-08-18
RFC: Resolve type info for fragment spreads

jonas:
Add document field to TypeInfo so it can resolve fragment definitions
and provide type information for fragment spreads.

---

Motivated by [this case in sangria-codegen](https://github.com/mediative/sangria-codegen/blob/1f36a8458329bbbb2e1b31099811316cbf48d052/sangria-codegen/src/main/scala/com.mediative.sangria.codegen/Importer.scala#L99-L105). This complicates downstreams users of `TypeInfo` by requiring the document AST instance up front. Another way to go would be for `TypeInfo` to keep a "document AST stack" and optionally resolve fragment spreads if a previous `TypeInfo.enter(ast.Document)` call was made.

### DIFF
--- a/src/main/scala/sangria/validation/TypeInfo.scala
+++ b/src/main/scala/sangria/validation/TypeInfo.scala
@@ -12,6 +12,7 @@ class TypeInfo(schema: Schema[_, _]) {
   private val inputTypeStack: ValidatorStack[Option[InputType[_]]] = ValidatorStack.empty
   private val fieldDefStack: ValidatorStack[Option[Field[_, _]]] = ValidatorStack.empty
   private val ancestorStack: ValidatorStack[ast.AstNode] = ValidatorStack.empty
+  private val documentStack: ValidatorStack[ast.Document] = ValidatorStack.empty
 
   var directive: Option[Directive] = None
   var enumValue: Option[EnumValue[_]] = None
@@ -23,6 +24,7 @@ class TypeInfo(schema: Schema[_, _]) {
   def inputType = inputTypeStack.headOption.flatten
   def fieldDef = fieldDefStack.headOption.flatten
   def ancestors: Seq[ast.AstNode] = ancestorStack.toSeq
+  def document = documentStack.headOption
 
   def withInputType(inputType: InputType[_]) = {
     inputTypeStack push Some(inputType)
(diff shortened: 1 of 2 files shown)


src/main/scala/sangria/validation/TypeInfo.scala
```diff
@@ -48,6 +48,11 @@ class TypeInfo(schema: Schema[_, _]) {
       case ast.OperationDefinition(ast.OperationType.Subscription, _, _, _, _, _, _, _) ⇒
         typeStack push schema.subscription
         pushParent()
+      case fs: ast.FragmentSpread ⇒
+        document.fragments.get(fs.name).foreach { fd =>
+          typeStack.push(schema.allTypes get fd.typeCondition.name)
+          pushParent()
+        }
```
jonas: Yes, good idea.

src/main/scala/sangria/validation/TypeInfo.scala
```diff
@@ -115,6 +120,11 @@ class TypeInfo(schema: Schema[_, _]) {
       case ast.OperationDefinition(ast.OperationType.Subscription, _, _, _, _, _, _, _) ⇒
         typeStack.pop()
         parentTypeStack.pop()
+      case fs: ast.FragmentSpread ⇒
+        document.fragments.get(fs.name).foreach { fs =>
+          typeStack.pop()
+          parentTypeStack.pop()
+        }
```
jonas: Made this unconditional.

jonas:
I've updated the PR to keep a document stack as proposed above, which also nicely reduces the size of the patch.

It makes sense to keep abstractions close to the reference implementation to ensure compliance, so I understand if this PR cannot be merged.

For the sangria-codegen project, the various visitor APIs didn't seem to be a good fit for generating a custom data structure for child nodes and doing some aggregation for the parent node. My understanding is that it would require to create a new visitor with state for each level. However, I am still getting familiar with Sangria's API so I might have missed something.

OlegIlyenko:
I also very sorry that it took me so long to come back to this PR :( I just revisited the PR and I think it would be a nice addition, thanks a lot for the work you've put into it and your patience 👍

jonas:
No problem. Thanks for merging.

Issue #277, opened 2017-08-18
Prepared queries without known variables


msolomon:
I'd like to be able to prepare (and ideally, run query reducers over) queries that incorporate variables before I actually have values for the variables.

For example, I'd like to prepare a query that requires a `$userId` variable, then subsequently execute that prepared query for any user ID passed in the `variables`.

The main use case is to support efficient execution[1] of [persisted queries](https://dev-blog.apollodata.com/persisted-graphql-queries-with-apollo-client-119fd7e6bba5). Sangria's existing Prepared Query support is great, but requires that variables are known at preparation time.

Combined with #195, Sangria's support for prepared/persisted queries would be great.

What do you think? The include/skip (and custom?) directives would need to be applied after variables are known, but I think that's the main hiccup.

--------------
[1] I'm generating implementations to fulfill GraphQL queries dynamically, so it would be much more efficient if we did this once per query (without variables filled in) instead of on every request.

OlegIlyenko:
Thanks for opening the issue and continuing the discussion on this topic. In fact, I already did a few experiments in this area after #195. It is a bit non-trivial to add this feature, but not impossible :)

One of the unclear points (and I would say the most important one) is how to handle the variables that might influence complexity. For example:

```graphql
query Foo($lim: Int!) {
  users(limit: $lim) {
    id
    name
  }
}
```

In absence of variables, `$lim` is unknown. `users` returns the list of users, so somehow we need to provide information about the worst-case scenario which is the max size of the user list that can possibly be returned. What do you think is the best approach to deal with this situation?

If we consider the subset of query reducers that are currently available out-of-the-box, then `measureComplexity`/`rejectComplexQueries` are the only one that influenced by this.

msolomon:
Yes, that makes sense--I'm not surprised you already had this idea!

Variables show up in two places: arguments to fields, and arguments to directives. I think the least API-breaking change is probably extending `sangria.schema.Args` to include the notion of args without known values.

If that was the main change and `QueryReducer`s could be run at preparation as well as at execution time, then authors of `QueryReducer`s would need to branch depending on if variables they care about are missing (during query preparation) or available (during query execution). It seems fairly natural to extend `Executor.prepare` to take a list of query reducers as an argument for running at preparation time instead of execution time.

Another option is to introduce another type of reducer that runs at preparation time instead of execution time and doesn't have access to variable values (only their types). This might be less compelling.

msolomon:
Updated #281

[...]

PR #281, opened 2017-08-31
Prepared queries without variables

msolomon:
This is a quick-and-dirty sketch for what #277 might look like. 

The basic approach is to add a new `Executor.prepare` method that doesn't accept variables, and add `preparedQueryReducers` to `Executor`.

Query reducers are run depending on the scenario:

* If we are preparing a query *without* known variables
  + Run `preparedQueryReducers` now and defer `queryReducers` to runtime
* If we are preparing a query *with* known variables
  + Run `preparedQueryReducers` and `queryReducers` now
* If we are executing an unprepared query:
  + Run `queryReducers`

The implementation is complete enough for my use case, but is not merge-ready.

What do you think about separating preparation-time from run-time like this?

### DIFF
--- a/src/main/scala/sangria/execution/DeprecationTracker.scala
+++ b/src/main/scala/sangria/execution/DeprecationTracker.scala
@@ -9,7 +9,7 @@ trait DeprecationTracker {
 
 object DeprecationTracker {
   val empty = NilDeprecationTracker
-  val print = PrintingDeprecationTracker
+  val print = new PrintingDeprecationTracker(println)
 }
 
 object NilDeprecationTracker extends DeprecationTracker {
@@ -17,10 +17,10 @@ object NilDeprecationTracker extends DeprecationTracker {
   def deprecatedEnumValueUsed[T, Ctx](enum: EnumType[T], value: T, userContext: Ctx) = ()
 }
 
-object PrintingDeprecationTracker extends DeprecationTracker {
+class PrintingDeprecationTracker(printline: String => Unit) extends DeprecationTracker {
   def deprecatedFieldUsed[Ctx](ctx: Context[Ctx, _]) =
-    println(s"Deprecated field '${ctx.parentType.name}.${ctx.field.name}' used at path '${ctx.path}'.")
+    printline(s"Deprecated field '${ctx.parentType.name}.${ctx.field.name}' used at path '${ctx.path}'.")
 
   def deprecatedEnumValueUsed[T, Ctx](enum: EnumType[T], value: T, userContext: Ctx) =
-    println(s"Deprecated enum value '$value' used of enum '${enum.name}'.")
+    printline(s"Deprecated enum value '$value' used of enum '${enum.name}'.")
 }
--- a/src/main/scala/sangria/execution/DeprecationTracker.scala
+++ b/src/main/scala/sangria/execution/DeprecationTracker.scala
@@ -9,7 +9,7 @@ trait DeprecationTracker {
 
 object DeprecationTracker {
   val empty = NilDeprecationTracker
-  val print = new PrintingDeprecationTracker(println)
+  val print = new LoggingDeprecationTracker(println)
 }
 
 object NilDeprecationTracker extends DeprecationTracker {
@@ -17,10 +17,10 @@ object NilDeprecationTracker extends DeprecationTracker {
   def deprecatedEnumValueUsed[T, Ctx](enum: EnumType[T], value: T, userContext: Ctx) = ()
 }
 
-class PrintingDeprecationTracker(printline: String => Unit) extends DeprecationTracker {
+class LoggingDeprecationTracker(logFn: String => Unit) extends DeprecationTracker {
   def deprecatedFieldUsed[Ctx](ctx: Context[Ctx, _]) =
-    printline(s"Deprecated field '${ctx.parentType.name}.${ctx.field.name}' used at path '${ctx.path}'.")
+    logFn(s"Deprecated field '${ctx.parentType.name}.${ctx.field.name}' used at path '${ctx.path}'.")
 
   def deprecatedEnumValueUsed[T, Ctx](enum: EnumType[T], value: T, userContext: Ctx) =
-    printline(s"Deprecated enum value '$value' used of enum '${enum.name}'.")
+    logFn(s"Deprecated enum value '$value' used of enum '${enum.name}'.")
 }
--- a/src/main/scala/sangria/execution/Executor.scala
+++ b/src/main/scala/sangria/execution/Executor.scala
@@ -7,8 +7,8 @@ import sangria.schema._
 import sangria.validation.QueryValidator
 import InputUnmarshaller.emptyMapVars
 import sangria.execution.deferred.DeferredResolver
-
 import scala.concurrent.{ExecutionContext, Future}
+import scala.sangria.execution.QueryReducerExecutor
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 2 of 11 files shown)


OlegIlyenko:
Thanks a lot for working on it!

In general, I like the approach. There is one concern I have with making this feature a part of `Executor.prepare`. I feel that in a long term it can be error-prone and a bit misleading. This breaks the contract of `Args` - in terms of type-safety this means that things that were non-optional suddenly will throw exceptions if used in particular way (in complexity function). I do think that the feature itself is quite valuable and this might be a suitable tradeoff. But I would like to make it a bit more obvious when query reducers are executed without variables. I also would like to decouple this feature from the `prepare`, i think they have different semantics and meant to be used for different use-cases.

What do you think about making small refactoring and extracting `reduceQuerySafe` away from the `Executor`? It is quite independent from everything else, so I think it can be easily extracted elsewhere (like a companion object of `QueryReducer` or companion object of  `Executor` or somewhere else). After this is done, we can still use it from `Executor`, as before. But then we can also provide a nice API to execute query reducers against query document and schema with or without variables. This will make it quite explicit and will not couple this functionality to execution in any way.

On a side note. Using `Args.empty` in case of empty variables can be quite restrictive. We can still discover arguments that are defined in the query itself. I think this will require additional refactoring and might be considered an improvement of this feature. So we can tackle it as a separate step in different PR.

src/main/scala/sangria/execution/QueryReducer.scala
```diff
@@ -38,6 +37,8 @@ trait QueryReducer[-Ctx, +Out] {
 }
 
 object QueryReducer {
+  type ArgumentValuesFn = (ExecutionPath, List[Argument[_]], Vector[ast.Argument]) => Try[Args]
```
msolomon: I found this helpful when reading the code

src/main/scala/sangria/execution/QueryReducerExecutor.scala
```diff
@@ -0,0 +1,179 @@
...
+import sangria.schema._
+import sangria.util.tag.@@
+import sangria.validation.QueryValidator
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success}
+
+object QueryReducerExecutor {
+
+  def reduceQueryWithoutVariables[Ctx, Root](
```
msolomon: used when variables aren't known

src/main/scala/sangria/execution/QueryReducerExecutor.scala
```diff
@@ -0,0 +1,179 @@
...
+
+      executionResult match {
+        case Success(future) ⇒ future
+        case Failure(error) ⇒ Future.failed(error)
+      }
+    }
+  }
+
+  /** Returns either new Ctx or future of it (with time measurement) */
+  def reduceQuery[Ctx, Root, Val](
```
msolomon: used when variables are known (prepare, execute)

msolomon:
Updated with your suggested model.

Some notes:

* Literal arguments defined in the query will now be discovered
* Added `QueryReducerExecutor` which holds static methods to run `QueryReducer`s over `Document`s and produce a user context
* `reduceQuerySafe` moved to `QueryReducerExecutor` and renamed to `reduceQuery`
* `reduceQuery` moved to `QueryReducerExecutor` and renamed to `reduceQueryUnsafe`

I also went through and added notes inline. I'd love to add inline Scaladocs, but for much of the existing code I don't feel qualified to describe the purpose and scope of each class.

I look forward to your comments!

src/main/scala/sangria/schema/package.scala
```diff
@@ -190,13 +190,13 @@ package object schema {
     description = Some("Directs the executor to include this field or fragment only when the `if` argument is true."),
     arguments = IfArg :: Nil,
     locations = Set(DirectiveLocation.Field, DirectiveLocation.FragmentSpread, DirectiveLocation.InlineFragment),
-    shouldInclude = ctx ⇒ ctx.arg(IfArg))
+    shouldInclude = ctx ⇒ ctx.args.argOpt(IfArg).getOrElse(true))
```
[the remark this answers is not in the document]
OlegIlyenko: yeah... I don't really like this, but I also don't see a better way at the moment. Maybe you can add a comment above to clarify why this code is like this and under which circumstances the argument would be empty. (the same below)

src/main/scala/sangria/execution/DeprecationTracker.scala
```diff
@@ -9,18 +9,18 @@ trait DeprecationTracker {
...
+  val print = new PrintingDeprecationTracker(println)
 }
 
 object NilDeprecationTracker extends DeprecationTracker {
   def deprecatedFieldUsed[Ctx](ctx: Context[Ctx, _]) = ()
   def deprecatedEnumValueUsed[T, Ctx](enum: EnumType[T], value: T, userContext: Ctx) = ()
 }
 
-object PrintingDeprecationTracker extends DeprecationTracker {
+class PrintingDeprecationTracker(printline: String => Unit) extends DeprecationTracker {
```
OlegIlyenko: 👍 Since it is now changed, maybe we can rename it a bit. I would suggest:

```scala
class LoggingDeprecationTracker(logFn: String ⇒ Unit)
```

src/main/scala/sangria/execution/DeprecationTracker.scala
```diff
@@ -9,18 +9,18 @@ trait DeprecationTracker {
 
 object DeprecationTracker {
   val empty = NilDeprecationTracker
-  val print = PrintingDeprecationTracker
+  val print = new PrintingDeprecationTracker(println)
```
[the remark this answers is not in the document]
OlegIlyenko: 👍

src/main/scala/sangria/execution/QueryReducerExecutor.scala
```diff
@@ -0,0 +1,179 @@
...
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success}
+
+object QueryReducerExecutor {
+
+  def reduceQueryWithoutVariables[Ctx, Root](
+    schema: Schema[Ctx, Root],
+    queryAst: ast.Document,
+    userContext: Ctx,
+    root: Root,
```
OlegIlyenko: You don't really need the root value, isn't it? (it's not used anywhere)

src/test/scala/sangria/execution/QueryReducerSpec.scala
```diff
@@ -789,4 +790,71 @@ class QueryReducerSpec extends WordSpec with Matchers with FutureResultSupport {
...
+            schema,
+            query,
+            Info(Seq(100)),
+            (),
+            reducer :: Nil
+          )
+          .await
+        assert(res._1.nums == Seq(100, 1, 1, 1, 1), "expected 4 nodes to be visited")
+      }
+    }
```
OlegIlyenko: Maybe you can also add extra tests with some existing query reduces and without variables? (`MeasureComplexity` in particular since it would be the main use-case, as far as I understand) You probably will need to define different schema just for this test because complexity function would have a different implementation

src/main/scala/sangria/execution/QueryReducerExecutor.scala
```diff
@@ -0,0 +1,179 @@
...
+    fields: CollectedFields,
+    userContext: Ctx)(implicit executionContext: ExecutionContext): Future[(Ctx, TimeMeasurement)] =
+    if (queryReducers.nonEmpty) {
+      val sw = StopWatch.start()
+      reduceQueryUnsafe(schema, fieldCollector, valueCollector, rootTpe, fields, queryReducers.toVector, userContext)
+        .map(_ → sw.stop)
+        .recover { case error: Throwable ⇒ throw QueryReducingError(error, exceptionHandler) }
+    } else Future.successful(userContext → TimeMeasurement.empty)
+
+  private def reduceQueryUnsafe[Ctx, Val](
```
OlegIlyenko: This probably was not the easiest part of the codebase to read through :) It is quite mutable and untyped. Not very proud of this code, but this approach helps to reduce the object allocation. The main idea was to reduce performance impact as much as possible.

src/main/scala/sangria/execution/Executor.scala
```diff
@@ -90,18 +84,20 @@ case class Executor[Ctx, Root](
...
-        tpe ← getOperationRootType(operation, queryAst.sourceMapper)
+        tpe ← Executor.getOperationRootType(schema, exceptionHandler, operation, queryAst.sourceMapper)
         fields ← fieldCollector.collectFields(ExecutionPath.empty, tpe, Vector(operation))
-      } yield reduceQuerySafe(fieldCollector, valueCollector, unmarshalledVariables, tpe, fields, userContext) match {
-        case fut: Future[(Ctx, TimeMeasurement) @unchecked] ⇒
-          scheme.flatMapFuture(fut)(c ⇒ executeOperation(queryAst, operationName, variables, um, operation, queryAst.sourceMapper, valueCollector,
-            fieldCollector, marshaller, unmarshalledVariables, tpe, fields, c._1, root, scheme, validationTiming, c._2))
-        case (ctx: Ctx @unchecked, timing: TimeMeasurement) ⇒
+      } yield {
+        val argumentValuesFn: QueryReducer.ArgumentValuesFn =
```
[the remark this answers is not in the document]
msolomon: Ah, yes. I used to inject this function, but later changed my mind

src/test/scala/sangria/execution/QueryReducerSpec.scala
msolomon: I added a test with MeasureComplexity, let me know if you had more in mind

msolomon:
Any thoughts on the latest updates?

msolomon:
@OlegIlyenko is there anything I can do to make this easier to review?

OlegIlyenko:
Hey Mike, sorry for the delay. I actually wanted to come back to it during the last week, but haven't managed it 😅 I already went quickly through it and it looks great! 👍 Just wanted to take some time give it a final thorough look before merging. I will try to do it during the next few days and then merge it.

src/main/scala/sangria/schema/package.scala
```diff
@@ -190,13 +190,19 @@ package object schema {
     description = Some("Directs the executor to include this field or fragment only when the `if` argument is true."),
     arguments = IfArg :: Nil,
     locations = Set(DirectiveLocation.Field, DirectiveLocation.FragmentSpread, DirectiveLocation.InlineFragment),
-    shouldInclude = ctx ⇒ ctx.arg(IfArg))
+    // if we don't know if we should include it, then we should include it:
+    // ValueCollector will fail before we get here if values must be known, such as when preparing or executing a query,
+    // but for e.g. running a QueryReducer without known variables, we must be conservative
```
OlegIlyenko: 👍 Thanks

PR #287, opened 2017-10-23
Update to SBT 1.0.2

nightscape:


OlegIlyenko:
👍  Thanks! Grad that all deps are now on 1.x

PR #291, opened 2017-10-31
Fix regression where variables aren't passed to query reducers

msolomon:
In #281 , unmarshalled variables stopped being passed from the executor to the query reducer executor. This means that variables and query reducers no longer work together--the unit tests show that they work independently, but there was no integration test for the two components.

This fixes it, and adds a regression test.


OlegIlyenko:
Thanks for a quick fix!

msolomon:
Thanks, incredible turnaround time!

PR #293, opened 2017-11-06
Add relOnly and relOnlyCaching methods to Fetcher

kuppuswamy:
Added `relOnly` and `relOnlyCaching` methods to Fetcher object, so that the `fetch` function argument won't be a must and the `fetchRel` function argument alone can be used for creating relation based fetchers.


OlegIlyenko:
Thanks, look like a good addition! A few times, I also found myself in a situations where I only could express relations.

Issue #308, opened 2017-12-04
Sangria 1.3.3 changelog doesn't mention two potential regressions


BjRo:
Hi @OlegIlyenko,

we've tried to upgrade to Sangria 1.3.3 and ran into two regressions. All in all they're not big and manageable on our side. 

I'm putting them down here, so that this doesn't get lost (and also to use the opportunity to get some input on those points from you). I hope that's fine for you!

## Executable Definitions
The first one is that there's a [new query validation rule](https://github.com/sangria-graphql/sangria/commit/b453be54c657429c073a1c25ffd83506229a3b4b), that doesn't seem to be mentioned in the `Changelog`. 

This broke most of our IDL based schema tests, since we're using `QueryValidator.allRules` to validate incoming schemas. Filtering out the `ExecutableDefinitions` rule, fixed those tests. 

In general what do you think would be a good approach to re-use rules for validating incoming schemas? My feeling is that were probably a bit too eager to re-use the existing validation rules for our IDL there and should be more selective in the future. Thoughts?

## Mandatory fields in input types
This one is a bit unfortunate for us. We basically re-use `input` types for declaratively defining resolvers (in lack of an actual construct that can be parameterised for defining directives).

```graphql
input IdlResolver1 @httpGet(url: "/profiles/profile/2") {}
```
After upgrading we saw several of the following errors:

```
[error]    sangria.parser.SyntaxError: Syntax error while parsing GraphQL query. Invalid input ""/test") {\n}", expected ValueConst, ArgumentConst, DirectiveConst or InputValueDefinition (line 7, column 32):
[error]    input MyResolver @httpGet(url: "/test") {
[error]                                   ^ (QueryParser.scala:433)
```

After staring at this a bit, we tracked the change in behaviour down to [this commit](https://github.com/sangria-graphql/sangria/commit/b8af9168b6252457f69e53c6c43560d24204099f#diff-9c1db5cb2ae322dd07ca363370f6c46cR239), where the `input` type definition was changed to have at minimum one field.

From 

```scala
 ~ InputValueDefinition.*
```
to 
```scala
~ InputValueDefinition.+
```

From a pure spec perspective I understand the change (input types without fields don't make a lot of sense). And I guess most of our time our resolvers will have fields (we use this for parametrizing the resolver)

```graphql
input UpdateSettingsResolver @httpPatch(
   url: "/rest/#{product}/settings/#{ctx.rid}", 
   body: "#{settings}") {
    product: Expression!
[...]

PR #317, opened 2017-12-29
Initialize symbols before checking their annotations.

### LINKED ISSUE #172: "Field list is empty" when only using @GraphQLField via deriveContextObjectType
Example:

``` scala
trait Query {
  val db: Our.Db
  import db._

  @GraphQLField
  def allUsers: Seq[User] = db.run(quote(query[User]))
}

object Query {
  case class Ctx(query: Query)
  implicit val UserOutType = deriveObjectType[Ctx, User]()
  implicit val queryType = deriveContextObjectType[Ctx, Query, Unit](_.query)
  val schema = Schema(queryType)
}
```

This works just fine if I use:

``` scala
implicit val queryType =
  deriveContextObjectType[Ctx, Query, Unit](_.query, IncludeMethods("allUsers"))
```

I've tried reading over the macro a time or two but the cause doesn't jump out at me.

Any ideas?

dragos:
The Scala compiler uses laziness extensively to deal with recursive types. Annotations are one instance where an uninitialized type will be incorrect, so we call `info` before checking method annotations inside macros. The proper method, initialize, is not available in the public API so I just called .info, which is what most places in the compiler do, anyway.

Fix #172

### DIFF
--- a/src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
+++ b/src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
@@ -130,7 +130,7 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
         val allFields = classFields ++ additionalFields(config)
 
         if (allFields.nonEmpty) Right(allFields)
-        else Left(List(c.enclosingPosition → "Field list is empty"))
+        else Left(List(c.enclosingPosition → s"$targetType: Field list is empty"))
       case errors ⇒ Left(errors)
     }
   }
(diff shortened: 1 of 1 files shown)


OlegIlyenko:
This is a great insight, thanks a lot!

dragos:
Thanks for the quick turnaround! Do you have an estimate for the next release/a release containing this fix? One of our customers hit this issue and it would be nice to have a simple upgrade path.

dragos:
Mid-January would be fine as a date, what I'm more concerned about is how easy it will be to upgrade given the other changes to go into 1.3.4. Are there any breaking changes?

In the meantime, maybe instead of a snapshot you could publish a milestone (a stable version number). If a bugfix release is easy to cut I'm happy to port my changes to the 1.3.4 tag (though I tend to agree that this is not a critical fix).

OlegIlyenko:
Thanks for the info! v1.3.4 would be primarily focused on:

* SDL-related syntax and semantic changes
* A small validation overhaul (new validations + improvements to some existing one)

v1.3.3 already introduced some backward incompatible changes in SDL syntax. They are documented in #308. Since then we [discussed it and addressed](https://github.com/facebook/graphql/pull/90#issuecomment-348976743) these concerns on a spec and reference implementation level. So one of the goals of v1.3.4 is to provide more backward compatibility in terms of SDL syntax with pre v1.3.3, So I don't expect any major breaking changes. 

Since it is not critical at the moment, I would suggest not to rush it for now (during holidays period) and proceed with original release plan.

Issue #318, opened 2017-12-29
DeriveObjectSetting for methods' arguments description


fehu:
`@GraphQLDescription` is not a universal solution to argument description problem because
1. one might not have access to the method's code,
2. annotation usage would create a dependency on _sangria_.

As you've explained in question #232, 
> it would be relatively hard to precisely identify exact argument you want to document

But it is still [solvable](https://github.com/fehu/sangria/commit/f6b32a41311b61f5ebf437d3fb8ac82416ba8129) except for overloaded methods.

Since, as I understand, such methods cannot be renamed using derivation settings, and the only way to avoid `java.lang.IllegalArgumentException: All fields within a Type should have unique names!` is to use `@GraphQLName` annotation - one is forced to either use the annotations or create a `schema.Field` manually.

For the not-overloaded methods, the arguments are uniquely identified by method's and argument's names and could be configured using derivation settings.

OlegIlyenko:
In general, I don't have anything against introducing support for arguments. AFAIR, I haven't added them in the first place because I thought that it would be a bit cumbersome to use them and I assumed that if one exposes methods with arguments, then these methods are specifically designed for GraphQL API. So one has full control over them. These assumptions might be wrong, of course.

If there is a compelling reason to support arguments via settings (and not only via annotations), then let's do it. Would you be interested in preparing a PR? 

The issue with method overloading is a general problem and, AFAIU, not related to argument documentation. I guess the current solution for resolving it would be an annotation-based approach.

fehu:
Yes, I can prepare a PR. I've added settings for specifying arguments' descriptions and default values; I'm just not sure that the names I chose are consistent with yours.

| Setting name | Description |
|------------------|----------------|
| `MethodArgumentDescription` | set description for a single argument |
| `MethodArgumentsDescription` | set descriptions for multiple arguments |
| `MethodArgumentDefault` | set default value for a single argument |
| `MethodArgument` | set description and default value for a single argument |

Would you like that I add more settings, like for renaming the arguments?

OlegIlyenko:
Looks great 👍 A setting for renaming an argument also sounds like a good idea.

PR #339, opened 2018-02-12
Allow to rename, describe and set default arguments using `DeriveObjectSetting`s.

fehu:
#318: Added the following `DeriveObjectSetting`s:

| Setting name | Description |
|------------------|----------------|
| `MethodArgumentRename` | change name of single argument |
| `MethodArgumentDescription` | set description for a single argument |
| `MethodArgumentsDescription` | set descriptions for multiple arguments |
| `MethodArgumentDefault` | set default value for a single argument |
| `MethodArgument` | set description and default value for a single argument |

### DIFF
--- a/src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
+++ b/src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
@@ -80,7 +80,7 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
             if (field.accessor)
               Nil → q"(c: sangria.schema.Context[$ctxType, $valType]) ⇒ ${if (useFn) q"valFn(c.ctx)" else q"c.value"}.${field.method.name}"
             else
-              fieldWithArguments(field, ctxType, valType, useFn)
+              fieldWithArguments(config, field, ctxType, valType, useFn)
 
           val fieldType = field.method.returnType
           val actualFieldType = findActualFieldType(fieldType)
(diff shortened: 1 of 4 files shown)


src/main/scala/sangria/macros/derive/DeriveObjectTypeMacro.scala
```diff
@@ -251,6 +256,32 @@ class DeriveObjectTypeMacro(context: blackbox.Context) extends {
...
+        .getOrElse(Left(List(
+          pos → s"Unknown argument '$argName' of method '$method'. Known arguments are: ${knownArguments.map(_.name.decodedName) mkString ", "}"
+        )))
+    }
+    def validateHasArgument(pos: Position, methodName: String, argName: String) = getArgument(pos, methodName, argName).right.map(_ ⇒ Nil).merge
+    def validateArgumentDefault(pos: Position, methodName: String, argName: String, default: Tree, argType: Type) =
+      getArgument(pos, methodName, argName).right.map{
+        case arg if argType <:< arg.typeSignature.resultType ⇒ Nil
+        case arg ⇒ List(
+          pos → s"Wrong type of default value '$default' for argument '$argName' of method '$methodName': expected '${arg.typeSignature.resultType}', but got '$argType'."
```
OlegIlyenko: I think we would need to rely on the standard set of validations for a default value type. It is based on `ArgumentType`, `ToInput` type-classes as well as some run-time validations. It still will provide the same compile-time guarantees. It's a minor thing, I will adjust it after the merge.

fehu:
Thank you for accepting my changes.

PR #342, opened 2018-02-17
Travis build matrix: add 'oraclejdk9'

sullis:


OlegIlyenko:
Thanks for updating the config. Looks like the build is failing for scala 2.11 and Java 9:

https://travis-ci.org/sangria-graphql/sangria/jobs/342814257#L675

Do you know the root cause of this issue? On a first sight looks like older versions of Scala API have compatibility issues with java9.

OlegIlyenko:
Just reproduced it locally as well. Looks like the Java 9 + Scala 2.11 combination does not really work: https://stackoverflow.com/questions/39791996/scala-reflect-internal-missingrequirementerror-object-java-lang-object-in-compi?rq=1

sullis:
I'll fix the build matrix.

.travis.yml
```diff
@@ -4,6 +4,12 @@ scala:
   - 2.11.11
 jdk:
   - oraclejdk8
+  - oraclejdk9
+
+matrix:
+  exclude:
+    - scala: 2.11.11
+    - jdk: oraclejdk9
```
OlegIlyenko: Looks like Travis now only tests a single combination: scala 2.12 + java 8

I think the right config would be (only one list element):

```yaml
matrix:
  exclude:
    - scala: 2.11.11
      jdk: oraclejdk9
```

OlegIlyenko:
Looks great! Thanks again :+1:

PR #345, opened 2018-03-09
Don't allow ObjectTypes derived from case classes to overwrite each other

simonsaffer:
If two or more case classes with the same name in different packages are used to derive `ObjectType`s Sangria will silently choose one of them and ignore the others. This PR makes it so that an exception is thrown if two `ObjectType`s are found with the same name but derived from different case classes.


OlegIlyenko:
Looks great! Thanks a lot for improving the DX and making the schema definition less error-prone :+1:

simonsaffer:
@OlegIlyenko Thanks for looking at this and getting it in so quickly!

PR #350, opened 2018-03-22
Transform enum values' names with a macro setting

fehu:
Added a derivation setting for transforming GraphQL enum values, similar to `TransformFieldNames`.
#### Use case
Since graphql has no definition scopes, the following scala enums will cause collision:
```scala
sealed trait Color
object Color {
  case object Red extends Color
  case object Blue extends Color
  case object Green extends Color
  // etc
}

sealed trait TrafficLight
object TrafficLight {
  case object Red extends TrafficLight
  case object Yellow extends TrafficLight
  case object Green extends TrafficLight
}
```

The solution might be adding a prefix to graphql enum values, renaming, for example, traffic light colors to `TrafficLightRed`, `TrafficLightYellow` and `TrafficLightGreen`.

------------------------

The proposed setting should allow to do it in bulk.

### DIFF
--- a/src/main/scala/sangria/macros/derive/DeriveEnumSetting.scala
+++ b/src/main/scala/sangria/macros/derive/DeriveEnumSetting.scala
@@ -13,3 +13,5 @@ case class RenameValue(value: String, graphqlName: String) extends DeriveEnumSet
 
 case class IncludeValues(values: String*) extends DeriveEnumSetting
 case class ExcludeValues(fieldNames: String*) extends DeriveEnumSetting
+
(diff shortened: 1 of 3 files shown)


OlegIlyenko:
Thanks a lot for working this feature! :+1: At the first glance it looks good to me, but I would like to give it nor in-depth review soon.

> the following scala enums will cause collision

I also wanted to mention that this scenario should work quite fine even without name transformation. Enum value names and type names do not collide with each other.

fehu:
Thanks for the tip, I thought that enum names would collide. Anyway, this setting might be useful in some other case.

PR #351, opened 2018-03-26
Improve query parsing performance

guymers:
Use the fact that `lineIdx` is a sorted array to more efficiently calculate the previous cursor and the number of cursors between it and 0. This gives an approximate 3.5x speedup when parsing the kitchen sink query.


OlegIlyenko:
Thanks for the working on this improvement! In gebneral the chnage looks good to me. :+1: 

From the very beginning I strived to keep the project build as simple as possible. This has a number of advantages and makes it easier for newcomers to learn and contribute to the project. So I have a concern about the introcuction of multi-module build.

I would suggest to keep thos PR simpler and focus only on the actual code chnages. At the momnet I would prefer to have a banchmark as a separate project. If the benchmark would become bigger and there would be a strong argumnet in favour of its inclusion in than main build, I think we can reonsider it and discuss it as a separate issue.

What do you thing about it? Would it be ok for you to just keep chnages in the `PositionTracking.scala`?

guymers:
Sure, if you don't want to a have a multi-module build. I added the benchmark module so the performance of the change could be verified. It is a separate commit so I can easily remove it.

PR #363, opened 2018-04-30
Fix BeforeFieldResult field name

Codier:
Did a code search and found no use for this field. So this should be a safe fix.


OlegIlyenko:
Thanks!

Issue #388, opened 2018-07-23
Missing subType check for AbstractTypes


NicolasRouquette:
Currently, `sangria.schema.TypeComparators.isSubType(schema, subType, superType)` returns false
if both `subType` and `superType` are `AbstractType`s (e.g., `Interface`).

This prevents us from using a GraphQL schema like this; based on a variant of the Stars Wars schema.

```
interface Character {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
}

interface Humanoid implements Character {}

type Human implements Humanoid {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  homePlanet: String
}

interface Droid implements Character {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  primaryFunction: String
  follows: Character # Note that this field is typed by an interface.
}

interface HumanoDroid implements Droid {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  primaryFunction: String
  follows: Humanoid # Note that this field overrides Droid.follows and is typed by an interface subtype
}

interface MonoDroid implements Droid {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  primaryFunction: String
  follows: Droid # Note that this field overrides Droid.follows and is typed by an interface subtype
}

type PrimaryDroid implements HumanoDroid {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  primaryFunction: String
  follows: Humanoid
}

type SecondaryDroid implements MonoDroid {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  primaryFunction: String
  follows: Droid
}
```

To demonstrate this, I implemented this example in a variant of the Sangria playground:

https://github.com/NicolasRouquette/sangria-playground/tree/MissingSubtypeCheckForAbstractTypeExample

Compile & run then go to: localhost:9000/render-schema
This will trigger an exception caused by this:

```
Caused by: sangria.schema.SchemaValidationException: Schema does not pass validation. Violations:

Droid.follows expects type 'Character', but SecondaryDroid.follows provides type 'Droid'.

Droid.follows expects type 'Character', but PrimaryDroid.follows provides type 'Humanoid'.
	at sangria.schema.SchemaValidationRule$.validateWithException(SchemaValidationRule.scala:42)
[...]

PR #389, opened 2018-07-23
Variant of starWars with interface subtyping failure

NicolasRouquette:
Demonstrates the problem (https://github.com/sangria-graphql/sangria/issues/388) on a StarWars2 unit test variant that fails.

### DIFF
--- a/src/main/scala/sangria/ast/QueryAst.scala
+++ b/src/main/scala/sangria/ast/QueryAst.scala
@@ -429,7 +429,7 @@ case class EnumTypeExtensionDefinition(
 
 case class UnionTypeExtensionDefinition(
     name: String,
-    types: Vector[NamedType],
+    interfaces_or_types: Vector[NamedType],
     directives: Vector[Directive] = Vector.empty,
     comments: Vector[Comment] = Vector.empty,
     location: Option[AstLocation] = None) extends TypeExtensionDefinition {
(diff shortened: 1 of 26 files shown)


NicolasRouquette:
Propose a conceptually simple strategy to fix #388:

Just like `InterfaceType[Ctx, Val]` has a field: `manualPossibleTypes`, 
the idea is to add an extra field: `manualPossibleInterfaces`.

NicolasRouquette:
@OlegIlyenko Can you please review the logic for interface subtyping in `sangria.validation.TypeComparators.isSubType`?

In particular, this line:

```
      case (t1: InterfaceType[_, _], t2: AbstractType) ⇒ (t1.name == t2.name) || schema.isPossibleInterface(t2.name, t1)
```

It uses a similar criteria than what's used for the `Named` case:

```
      case (t1: Named, t2: Named) ⇒ t1.name == t2.name
```

My original fix had a stronger criteria:

```
      case (t1: InterfaceType[_, _], t2: AbstractType) ⇒ (t1 == t2) || schema.isPossibleInterface(t2.name, t1)
```

However, this led to a failure in the schema extension test; see:
https://travis-ci.org/sangria-graphql/sangria/jobs/407010037#L1963

Since the spec says that type references are by name (see: http://facebook.github.io/graphql/June2018/#sec-Type-References) then it should be OK to use a name equality test to confirm that two types refer to the same thing.

NicolasRouquette:
This PR is out of scope of the [June 2018 Graph QL Spec](http://facebook.github.io/graphql/June2018/#sec-Interfaces)

PR #403, opened 2018-09-17
Format code using Scalafmt

nightscape:
This PR introduces Scalafmt as code formatter in the first commit and applies is in the second commit.
The configuration tries to be Git-friendly by [minimizing unnecessary diffs](https://scalameta.org/scalafmt/docs/faq.html#which-configuration-options-minimize-diffs-conflicts-in-version-control).
Of course the configuration is up for discussion, and if you have any wishes I can try to incorporate them.


OlegIlyenko:
Thank you for this suggestion. I normally avoid taking a strong reluctant position on PRs, but I afraid I feel quite uncomfortable introducing scalafmt. I don't see much benefit in it, quite in contrary, it will mess up the history and has some other negative side-effect in my opinion.

On a general note, I'm not a fan of automated code formatters. From what I have seen, they produce bad to mediocre results in terms of the actual formatting, regardless of the configuration. It is a bit bigger topic and it would be hard to describe it in one comment, but I also believe there is more more to this topic (technical as well as non-technical reasons).

I will close this PR, at least for now. I don't want to stay closed to this idea entirely, but at the moment I'm unconvinced that introduction of automated formatting tool would be overall beneficial.

nightscape:
I've been using Scalafmt on several projects and so far I've been very happy with the results.
If you haven't tried it you should give it a go.
I do understand your concern with the massive history changes.
A different approach to this would be to configure formatting such that it only gets applied to changed files or even changed lines (Intellij IDEA is able to do that).
For me, having an automated code formatter configured to my style preferences makes my life as project maintainer much easier because I don't have to remind contributors to please format the code like the rest of the code base. But as always ymmv.

PR #404, opened 2018-09-17
Assorted refactorings

nightscape:
I'm trying to understand the code a little better and applying small refactorings here and there.
I'm grouping them by theme, so that changes you don't like can be taken out.
Any comments welcome :smiley:

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/deferred/Fetcher.scala
+++ b/modules/core/src/main/scala/sangria/execution/deferred/Fetcher.scala
@@ -62,13 +62,7 @@ class Fetcher[Ctx, Res, RelRes, Id](
     val allIds =  MutableMap[Relation[Any, Any, Any], MutableSet[Any]]()
 
     def addToSet(rel: Relation[Any, Any, Any], id: Any) =
-      allIds.get(rel) match {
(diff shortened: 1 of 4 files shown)


travisbrown:
@nightscape Would you be interested in [migrating this PR to the new GitHub organisation](https://github.com/sangria-graphql/sangria/issues/463)?

nickhudkins:
Hey @nightscape I've rebased and broken out the commits that Oleg had mentioned being worried about (with regards to performance). 

My inclination is to drop the commits with `[WARNING]` in them unless we had some sort of benchmark to know if it has impact, even though my gut says that the changes don't scream performance issues.

I'm gonna go ahead and leave this PR open for now as is, and if I haven't heard back by early next week I'll go ahead and split the `WARNING` commits into their own PR.

nightscape:
Hi @nickhudkins, thanks for taking care!
I wouldn't expect performance issues either (rather the other way around as the Scala collections might have optimized implementations), but your proposal about splitting into two PRs sounds like a good approach.

nickhudkins:
@nightscape split :)

yanns (approved):
👍

PR #412, opened 2018-10-07
Support for repeatable directives

### LINKED ISSUE #409: Support for repeatable directives
https://github.com/graphql/graphql-js/pull/1541
https://github.com/facebook/graphql/pull/472

We also need to make sure that unique directives are properly validated: https://github.com/facebook/graphql/issues/429#issuecomment-392990065

OlegIlyenko:
Merge as soon as https://github.com/graphql/graphql-js/pull/1541 is merged.

Closes #409


spawnia:
This should be ready as https://github.com/graphql/graphql-js/pull/1965 has landed in the reference implementation.

spawnia:
I think this is a nice feature, but have never written a single line of Scala code in my life. I hope someone else gets to it - godspeed!

yanns (approved):
The code is looking good for me.
I personally don't use custom directives and I cannot judge on how useful this feature is.
Anyone here who would need that?

nikola-mladenovic:
@yanns since the https://www.apollographql.com/docs/federation/federation-spec/#federation-schema-specification contains repeatable directives, this might be the right motivation to merge this work?

yanns:
Repeatable directives are now in the spec: http://spec.graphql.org/October2021/#sec-Type-System.Directives

yanns:
As there are many conflicts to fix in this PR, I've opened a new one with about the same changes: https://github.com/sangria-graphql/sangria/pull/899

PR #420, opened 2018-11-21
Use AstSchemaBuilder from args rather the default one

Axxiss:
On the current implementation `definitions(document, builder)` doesn't use the builder argument at all  and instead always use `AstSchemaBuilder.default`. So this PR make use of the custom builder and pass to the `AstSchemaMaterializer` as an argument.

``` scala
 // This implementation provides the default AstSchemaBuilder
  def definitions(document: ast.Document): Vector[Named] =
    definitions[Any](document, AstSchemaBuilder.default)

 // Here we accept a custom AstSchemaBuilder as an argument but we don't use it. We use the default one instead
  def definitions[Ctx](document: ast.Document, builder: AstSchemaBuilder[Ctx]): Vector[Named] =
    new AstSchemaMaterializer[Ctx](document, AstSchemaBuilder.default).definitions
```

### DIFF
--- a/src/main/scala/sangria/schema/AstSchemaMaterializer.scala
+++ b/src/main/scala/sangria/schema/AstSchemaMaterializer.scala
@@ -699,7 +699,7 @@ object AstSchemaMaterializer {
     definitions[Any](document, AstSchemaBuilder.default)
 
   def definitions[Ctx](document: ast.Document, builder: AstSchemaBuilder[Ctx]): Vector[Named] =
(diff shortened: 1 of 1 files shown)


travisbrown:
@Axxiss This change looks reasonable to me. Would you be interested in [migrating this PR to the new GitHub organisation](https://github.com/sangria-graphql/sangria/issues/463) for review?

Axxiss:
Moving the PR to the new orga 👍

Issue #418, opened 2018-11-12
Schema transformation


OlegIlyenko:
The idea is to provide an easy and convenient way to transform existing schema. Since all types are cross-referenced within the schema, the whole schema needs to be re-created in a similar way as in the [AST materializer](https://github.com/sangria-graphql/sangria/blob/master/src/main/scala/sangria/schema/AstSchemaMaterializer.scala) where it re-creates existing types with the help of the builder.

In this case a dedicated builder might be useful as a low-level implementation detail. But it also would be nice to have a high-level DSL. Here is just an idea of how it might look like:

```scala
val newSchema =
  schema
    .mapObjectType("Book", _.rename("Publication"))
    .mapOutputTypes(_.filterFileds(field ⇒ !field.tags.contains(Preview)))
```

Axxiss:
Hey @OlegIlyenko I'm taking a look at this issue and would like to double check that the approach we use internally fits what you have in mind.

This is our current implementation

``` scala
class PreviewSchemaFilter extends DefaultAstSchemaBuilder[RequestContext] {

  override def extendObjectType(origin: MatOrigin,
                                existing: ObjectType[RequestContext, _],
                                extensions: List[ast.ObjectTypeExtensionDefinition],
                                fields: () => List[Field[RequestContext, Any]],
                                interfaces: List[InterfaceType[RequestContext, Any]],
                                mat: AstSchemaMaterializer[RequestContext]): ObjectType[RequestContext, Any] = {
    existing
      .asInstanceOf[ObjectType[RequestContext, Any]]
      .copy(
        interfaces = interfaces,
        fieldsFn = () => fields().filterNot(_.tags.contains(Preview))
      )
  }
}

object PreviewSchemaFilter {

  def filter(sangriaSchema: SangriaSchema): SangriaSchema = {
    val emptyExtension = ObjectTypeExtensionDefinition(sangriaSchema.query.name, Vector.empty, Vector.empty)
    AstSchemaMaterializer.extendSchema(sangriaSchema, Document(Vector(emptyExtension)), new PreviewSchemaFilter)
  }
}
```

In a nutshell, we extend an existing schema and filter our the fields that have a certain flag. This allows us to create two schema and use one or the other depending on the incoming request.

So if this looks good, I would use this as a starting point for a low-level implementation and move towards a more generic approach.

OlegIlyenko:
[...]

PR #421, opened 2018-11-21
Extract extension methods from AstSchemaBuilder 

Axxiss:
As discussed https://github.com/sangria-graphql/sangria/issues/418#issuecomment-440458452 this PR aims to settle the ground for decoupling schema extension from  materialization.

Creates a new trait `AstSchemaExtender` which contains all schema extension methods definition as originally defined in `AstSchemaBuilder`.

Furthermore, it creates a new default `AstSchemaExtender` implementation moving the implementation from `DefaultAstSchemaBuilder` to `DefaultAstSchemaExtender`

In order to keep backwards compatibility  `AstSchemaBuilder` signature remains unchanged  with the following tradeoffs
- `AstSchemaBuilder` extends  `AstSchemaExtender`
- `DefaultAstSchemaBuilder` extends `DefaultAstSchemaExtender`

### DIFF
--- a/src/main/scala/sangria/schema/AstSchemaBuilder.scala
+++ b/src/main/scala/sangria/schema/AstSchemaBuilder.scala
@@ -1,15 +1,15 @@
 package sangria.schema
 
 import language.existentials
-
 import sangria.ast
 import sangria.execution.FieldTag
 import sangria.marshalling.{FromInput, MarshallerCapability, ScalarValueInfo, ToInput}
+import sangria.schema.extension.AstSchemaExtender
 import sangria.validation.Violation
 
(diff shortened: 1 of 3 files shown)


travisbrown:
@Axxiss Would you want to make a case for this change going into 2.0? We're hoping to [finalise what will be in it](https://github.com/sangria-graphql/sangria/issues/463) this week.

Axxiss:
@travisbrown given the current situation with sangria. As this PR only moves code around without adding any value to end users, I think we can leave this out for now and wait for the next iteration. 

If the new organization still want to move forward with #418 I will create the PR in the new repo.

Issue #427, opened 2018-12-16
Provide query arguments alongside name projections


robconrad:
In [this](https://stackoverflow.com/questions/41325218/how-to-know-requested-fields-in-resolver-of-object-in-sangria-graphql) StackOverflow comment it is recommended to use built-in projection support to avoid the N+1 query problem when resolving nested objects. This works perfectly if the only requirement of the generated query is knowing which items to select since those are all available as `ProjectedName` vectors. However, if arguments can exist within the nested objects that are required to form predicates, sorts, or other non-shape parameters of the query it becomes necessary to dive into the ast to pull the arguments out.

Would it be possible to provide the args at the projected level of the ast the `ProjectedName` represents along with the name?

nickhudkins:
#563

PR #430, opened 2018-12-24
extend ProjectionNames to include Args

robconrad:
This PR extends the Projector facility with access to the Args present at each projection node. This enables use cases for n+1 query mitigation that require access to arguments in order to properly formulate the query.

I bumped the version by one patch increment, but this change may in fact require a minor version bump because anyone relying on case comprehension on the `ProjectedName` class will have their code broken if they pick up this build. 

Addresses https://github.com/sangria-graphql/sangria/issues/427

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 name := "sangria"
 organization := "org.sangria-graphql"
-version := "1.4.3-SNAPSHOT"
+version := "1.4.4-SNAPSHOT"
 
 description := "Scala GraphQL implementation"
 homepage := Some(url("http://sangria-graphql.org"))
--- a/src/main/scala/sangria/execution/Resolver.scala
+++ b/src/main/scala/sangria/execution/Resolver.scala
@@ -1165,7 +1165,7 @@ class Resolver[Ctx](
                       else Vector(field.name)
 
                     projectedName.map (name ⇒
-                      ProjectedName(name, loop(path.add(astField, objTpe), field.fieldType, fields, currLevel + 1)))
+                      ProjectedName(name, loop(path.add(astField, objTpe), field.fieldType, fields, currLevel + 1), Args(field, astField)))
                 }
                 .flatten
             case Failure(_) ⇒ Vector.empty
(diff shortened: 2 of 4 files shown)


build.sbt
```diff
@@ -1,6 +1,6 @@
 name := "sangria"
 organization := "org.sangria-graphql"
-version := "1.4.3-SNAPSHOT"
+version := "1.4.4-SNAPSHOT"
```
robconrad: Should this be bumped to `1.5.0-SNAPSHOT` since the signature of the case class `ProjectedName` changed, which is potentially a breaking API change for users of that class?

paulpdaniels:
Is the args object useable in that fashion? The ast argument is not the same as the `schema.Argument`. For instance it isn't useable with `withArgs`. How will these actually be used in a resolver?

robconrad:
Absolutely, I am using this code already in production. When you wrap your resolver in `Projector` you get access to `projectedNames: Vector[ProjectedName]`, each of which then (with this PR) provides not only the name of the projected child node, but also the (schema) args that were used on that child.

I'm not sure why you'd want the ast args in the resolver.. Are you referencing the `withArgs` methods on the `WithArguments` trait? If so I believe those are schema args, not ast args.

paulpdaniels:
Yeah, I'm just confused about what you would do with ast level arguments. With the schema level ones you can easily extract and case match. It seems like it would be more work to do it with the ast versions no?

robconrad:
These are the schema level arguments.. perhaps you are being misled by the constructor `Args(field, astField)`? That's a schema args constructor that takes an ast `WithArguments` element. Or maybe I am totally missing something obvious..

kirach:
@robconrad given the new (temporary) organization for sangria - @sangria-graphql-org and https://github.com/sangria-graphql/sangria/issues/446#issuecomment-546281588, do you mind re-creating this PR in the new repo (https://github.com/sangria-graphql-org/sangria)? This is a very important change and we've been successfully using it in production so I think it should be part of sangria. I can re-create this PR by myself but don't want to take credit for your work.

travisbrown:
@kirach Are you interested in taking over this change and migrating it to the new org?

kirach:
@travisbrown yep I'm on it, will do it later today or tomorrow

kirach:
Moved to the new repo here: https://github.com/sangria-graphql-org/sangria/pull/10

PR #441, opened 2019-04-23
Apply max batch size to relation

maxperrimond:
Tried to apply max batch size from the fetcher config to relations. Same principle as `resolveEntities` by grouping `nonCachedIds` by the size but for each relation and then merging them back.


travisbrown:
@maxperrimond Would you be interested in [migrating this PR to the new GitHub organisation](https://github.com/sangria-graphql/sangria/issues/463)?

maxperrimond:
@travisbrown sure

PR #448, opened 2019-06-20
Add update cache methods to Fetcher

maxperrimond:


nickhudkins:
Hi @maxperrimond, it's been a while here, so my apologies.

**Two questions:**
  1.) Are you still actively using Sangria?
  2.) Do you have interest in moving this along?

One thing I'd like to understand a bit more is the use case. Could you provide some motivation behind why this change is important, specifically what it allows a user of Sangria to do that is not possible today?

My immediate reaction is that this is highly dependent on the fetcher cache existing ACROSS requests to provide benefit, but it is entirely possible that I am just missing the intention.

Thanks!
Nick

avelx:
Assuming there is no response to the questions raised, should it be fine to close this PR?

maxperrimond:
Hi @avelx, sorry I couldn't respond to the previous questions.

Currently I'm no longer using Sangria or Scala, so this change is no longer relevant for me.
For context, I don't remember exactly, but it was an edge case anyway.

If there's no need for this addition, I will close this PR.

PR #451, opened 2019-07-17
[WIP] Scala 2.13

jtjeferreira:
fixes #440 

most changes were related to collections, but the one that surprised me the most was https://github.com/scala/scala/pull/6589

PR #453, opened 2019-08-06
Updates for 2.13.0 [WIP]

travisbrown:
Update: I went ahead and added [MiMa](https://github.com/lightbend/mima) to the following Scala 2.13 PRs so that I could check binary compatibility for 2.11 and 2.12, and I've updated versions, etc. as necessary:

* 1.0.0 → 1.0.1: https://github.com/sangria-graphql/sangria-streaming-api/pull/3
* 1.0.3 → 1.0.4: https://github.com/sangria-graphql/sangria-marshalling-api/pull/5
* 1.0.1 → 1.0.2: https://github.com/sangria-graphql/sangria-marshalling-testkit/pull/3
* 1.0.1 → 1.0.2: https://github.com/sangria-graphql/sangria-spray-json/pull/2
* 1.0.0 → 1.0.1: https://github.com/sangria-graphql/sangria-argonaut/pull/2
* 1.2.1 → 1.3.0: https://github.com/sangria-graphql/sangria-circe/pull/19
* 1.0.0 → 2.0.0: https://github.com/sangria-graphql/sangria-monix/pull/2
* 1.0.0 → 2.0.0: https://github.com/sangria-graphql/sangria-ion/pull/1

This PR also depends on one additional PR and update where I'm not checking MiMa, but it's a macro dependency with trivial changes:

* 0.1.1 → 0.1.2: https://github.com/sangria-graphql/macro-visit/pull/3

MiMa does report that this PR breaks binary compatibility with 1.4.2, but at a glance it looks like most of that breakage was already in the current master. I've updated the version to 2.0.0 to reflect the current bincompat status.

---

This PR includes @jtjeferreira's commits from #451 but continues the work of upgrading by adding several other changes:

* In one place we're running into an 2.13 regression (?) where an implicit overload is no longer resolved correctly. I've reported this as a compiler bug [here](https://github.com/scala/bug/issues/11662), but there's [an easy workaround](https://github.com/sangria-graphql/sangria/commit/7b4f816e50f2bc449932c908b07b682c80c9f44c) (make both `apply`s non-implicit, and add a new implicit `convert` method with the same signature and implementation).
* There's one other place where we run into new restrictions on overloads, but I've just [changed the tests there](https://github.com/sangria-graphql/sangria/commit/f59ff564fdc39729c20c9440bd68c00b9eeb8e50). I'm not sure how common this usage is in the wild, or if it's worth trying to support on 2.13. In any case the workaround is pretty straightforward.
* I've temporarily removed the `StreamSpec` tests because neither Monix or RxScala have 2.13 releases. Before this PR is merged we should reinstate these tests for 2.11 and 2.12—I've not done that yet because I don't have locally published snapshots of the Sangria modules for these libraries. (Update: Monix 3.0.0 is out and I've reinstated the sangria-monix streaming tests.)
* I've replaced the Monix observable-based `SubscriptionStream` in the `BatchExecutor` tests with a simple fs2-based version. Since these tests are meant to exercise `BatchExecutor`, the implementation shouldn't really matter.
* I've replaced one instance of Monix's `AtomicInt` with the Java `AtomicInteger`, since I don't see any reason to use the Monix one.
* Some query parsing tests were broken because of error formatting changes in Parboiled2 between 2.1.4 and 2.1.7. I've fixed the tests to match the new errors.
* Two test suites included tests that depended on map traversal order, which in some cases changed between 2.11 / 2.12 and 2.13 (for a minimal example try `List(1, 2).groupBy(identity).map(_._2.head)`). I've just sorted before comparing or in one case added a new recursive unordered comparison method.

With these changes all tests pass (assuming you've published all the dependency snapshots locally).


build.sbt
```diff
@@ -6,51 +6,50 @@ description := "Scala GraphQL implementation"
...
-  "org.parboiled" %% "parboiled" % "2.1.4",
+  "org.parboiled" %% "parboiled" % "2.1.8",
 
   // AST Visitor
-  "org.sangria-graphql" %% "macro-visit" % "0.1.1",
+  "org.sangria-graphql" %% "macro-visit" % "0.1.2-SNAPSHOT",
 
   // Marshalling
-  "org.sangria-graphql" %% "sangria-marshalling-api" % "1.0.3",
+  "org.sangria-graphql" %% "sangria-marshalling-api" % "2.0.0-SNAPSHOT",
```
jtjeferreira: as I commented in https://github.com/sangria-graphql/sangria-marshalling-api/pull/4#issuecomment-511390492 I think we can make avoid breaking compatibility and use `1.0.4` instead...

travisbrown: +1 (for now I've just gone with the versions in the existing PRs for these dependencies).

travisbrown:
For what it's worth my initial goal for the changes in this PR was to get Sangria published internally so I could evaluate our 2.13 upgrade, and I can confirm that it seems to work (at least to the degree that our tests are passing).

build.sbt
ingemaradahl: I don't see the harm in breaking binary compatibility in `sangria-marshalling-api` since there is work required in all marshalling glue libraries anyway. Personally I don't like using deprecated stuff when there's a chance to fix it.

jeffrey-aguilera:
99.9% of this PR is related to renaming arrows and its stated purpose is buried among 179 changed files. Perhaps it should be split?

mjadczak:
I think the arrow changes are very clearly contained in a single separate commit - in fact the rest of the changes are also well organised in commits.

travisbrown:
Now that [Monix 3.0.0](https://github.com/monix/monix/releases/tag/v3.0.0) is out and available for Scala 2.13.0 I've reinstated the Monix part of `StreamSpec`, which shrinks the diff here a little. (Note that this depends on [this sangria-monix PR](https://github.com/sangria-graphql/sangria-monix/pull/2).)

travisbrown:
@jeffrey-aguilera I could split up the PR to exclude @jtjeferreira's 9f5f31667be055a49ec9b48c9459f4d9a, but that results in a lot of deprecation warnings on 2.13 that swamp the build, and it's straightforward to review the diff since then in a terminal, so I'd prefer to leave it here.

build.sbt
travisbrown: @ingemaradahl In Scala 2.13 `Traversable` is a type alias for `Iterable`, so even if we keep using it there, that doesn't impose any deprecation cost downstream (users can just treat the return value of `getMapKeys` as an `Iterable` on 2.13).

travisbrown:
Superseded by this PR: https://github.com/sangria-graphql-org/sangria/pull/1

PR #454, opened 2019-08-08
support for configuring max concurrent batches in fetcher

vivekdhayaal:
- added a new config param maxConcurrentBatches to configure
the max number of batched IDs that can be fetched concurrently by fetcher.
- not specifying any value for this param will fallback to the existing
behaviour of running all batches concurrently.


nickhudkins:
superseded by #612

PR #457, opened 2019-09-11
Correctly resolve member types of type-parameterized classes

### LINKED ISSUE #261: Use type arguments in deriveObjectType
I try to return a value of

    case class MyValue[A](value: Option[A])

When I try to derive it from using a macro

    deriveObjectType[Unit, MyValue[Double]]()

it fails withe compile error: 

    Error:(25, 98) Can't find suitable GraphQL output type for Option[A]. If you have defined it already, please consider making it implicit and ensure that it's available in the scope.

argson67:
For macro-based schema derivation
Resolves #261


yanns:
Could you please rebase on master?

nickhudkins:
@yanns can we close this PR since #527 has been merged? I am also happy to do this sort of stuff if you're ok with making me a contributor :)

Issue #296, opened 2017-11-20
Optimize OverlappingFieldsCanBeMerged


objmagic:
`QueryReducer.rejectComplexQueries` become stuck with a 412KBytes GraphQL query selecting ~150000 fields. Did a profiling and stacktrace shows we are doing tremendous amount of work in `OverlappingFieldsCanBeMerged`

An example of the query is at [here](https://gist.github.com/objmagic/3c881449fcdb3a812a371b86bfa5a3c9)

We are using:

* **Sangria version 1.3.2**
* **JVM 8 (with some Twitter internal customization)**

A screenshot of part of the stacktrace (sorry I cannot provide complete stacktrace).

![image](https://user-images.githubusercontent.com/3656079/33035178-14f4d5e4-cddf-11e7-8077-a087137da451.png)

objmagic:
@OlegIlyenko 

Thanks a lot for your prompt reply. I would love to work through this issue with you.

Let me see if I can get a better stacktrace.

OlegIlyenko:
This would be awesome! 👍 As a first step, I think it would be really helpful to verify that mentioned validation is indeed the bottleneck. Can you please exclude it in your tests:

```scala
QueryReducerExecutor.reduceQueryWithoutVariables(
  ...
  queryReducers = ...,
  queryValidator = 
    new RuleBasedQueryValidator(
      QueryValidator.allRules
        .filterNot(_.isInstanceOf[OverlappingFieldsCanBeMerged])))

// or

Executor.execute(
  ...
  queryReducers = ...,
  queryValidator = 
    new RuleBasedQueryValidator(
      QueryValidator.allRules
        .filterNot(_.isInstanceOf[OverlappingFieldsCanBeMerged])))
```

If we can verify that it cases the slowdown, then we can focus on `OverlappingFieldsCanBeMerged`.

objmagic:
remove the `OverlappingFieldsCanBeMerged` validator solves the issue.

OlegIlyenko:
Great, this means that we can isolate the issue and focus the effort on optimising the `OverlappingFieldsCanBeMerged` validation. Maybe you could also change the issue title to reflect this fact?

If you would like to investigate the issue as well, this here are some usefule resources:

* The validation implementation: [OverlappingFieldsCanBeMerged](https://github.com/sangria-graphql/sangria/blob/master/src/main/scala/sangria/validation/rules/OverlappingFieldsCanBeMerged.scala)
[...]

PR #458, opened 2019-09-18
Add a faster version of the field selection merging validation

SimonAdameit:
The current OverlappingFieldsCanBeMerged validation has performance problems with a high number of fragments or selection sets. This implements a faster version of OverlappingFieldsCanBeMerged, that can deal with a high number of fragments or selection sets without a blowup in runtime.

It also includes JMH for benchmarks.

Fixes: https://github.com/sangria-graphql/sangria/issues/296

The algorithm is described in:
https://tech.xing.com/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01

Below is an example benchmark graph showing the performance improvement for multiple fragments in the same selection set.
2019-09-16 NoOverlapFrag

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,4 @@
+template: |
+  ## What’s Changed
+
+  $CHANGES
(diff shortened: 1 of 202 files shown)


SimonAdameit:
Migrated to the new temporary sangria organization: https://github.com/sangria-graphql-org/sangria/pull/12

PR #468, opened 2019-12-23
Args variables

kirach:


yanns:
Was already tried & reverted:
https://github.com/sangria-graphql-org/sangria/pull/10#issuecomment-571471876

dragisak:
What's the status of this PR ? Is it waiting for a patch from @kirach ?

kirach:
Sorry guys, it's been on my todo list forever. We have a fork of sangria with the proper fix that we've been using for quite some time and it looks to be working fine. I just need to find some time to re-create another PR with the correct fix here.

cc @dragisak @yanns

nickhudkins:
Hey @kirach, I am happy to do the work if you don't have time. Is the working fork `dv01-inc/sangria`? If so I can rebase and get this tidied up. @yanns it looks like you had a application where this change exposed a defect, any chance you could explain how we could get a failing test case to ensure we don't accidentally re-introduce that?

yanns:
@nickhudkins I don't have any minimal test for it.

ibsmfar:
Hi! What's the status on this?

nickhudkins:
Hey @ibsmfar, while I am happy to work on this, this had been fixed in a PR, but then reverted due to an issue here:

https://github.com/sangria-graphql-org/sangria/pull/10#issuecomment-571471876

Unfortunately I don't know what it breaks, and we don't have a test that would show what breaks. I'll get this PR rebased, and maybe we could get a test release out the door to see if the breakage was elsewhere and since resolved.

@yanns you had commented that it broke a production application, was this an application you worked on? Is it a public repo that I could dig around in at least?

If not, I'll see if I can reproduce just by looking at the stack trace and query and seeing if that helps me get anywhere.

yanns:
> @yanns you had commented that it broke a production application, was this an application you worked on? Is it a public repo that I could dig around in at least?

It's not public.

nickhudkins:
#563 supersedes this PR

Issue #346, opened 2018-03-12
Add a new annotation to specify ID type for derive macros


athanikkal:
Currently derive macros always create StringType for Scala strings fields in the case class. It would be really helpful to have an annotation to indicate that a field is an IDType and then the derived ObjectType or InputObjectType should honor this.

OlegIlyenko:
I agree, it might be a nice feature. We can even try to make it more generic and allow to override any field type with the help of annotation and/or macro setting.

tuleism:
Hi, I'm looking into working on this. Can you give some examples how it's gonna be used? Here's what I have in mind:

- Override field type with annotation: we would need 2 annotations, one for ObjectType and one for InputObjectType.
- Override field type with macro setting: change type for fields having name or type satisfying a predicate. Is this actually useful?

athanikkal:
I am sorry that I didn't get back to you earlier, somehow I missed the github notification.

The way I am planning to use is by specifying annotations for ObjectType and InputObectType at the field level.  I don't have a use case for the second option above at the moment, but I would think it might be useful in some cases.

ErikHedblom:
In addition to changing strings into `IDType` I think one of the major use cases would be to set  the type of a filed to `InterfaceType` or `UnionType`.

ErikHedblom:
Also I think it would be very useful if this feature can handle compound types like `OptionInputType(ListInputType(IDType))`

DuncanHills:
This can be closed now that the GraphqlInputType/GraphQLOutputType annotations exist right?

sh0hei:
@DuncanHills I think it can be closed, Thanks.

PR #472, opened 2020-01-21
GraphQLOutputType and GraphQLInputType annotations

khoberg:
Added macro support for 2 new annotations: GraphQLOutputType and GraphQLInputType
These annotations take a Sangria OutputType or InputType respectively, and uses that as the field type instead of the found implicit type.

This was requested in #346


khoberg:
Re-creating against sangria-graphql-org/sangria

PR #516, opened 2020-09-29
Fix broken website links in README

anagrius:
The links are still to the old dead website.

Unfortunately we don't have "Try Sangria" running anywhere, so the link is updated to point at
the playground repo so people can run it themselves.


yanns (approved):

yanns:
thx a lot!

PR #520, opened 2020-10-06
parboiled 2.2.1

agourlay:
I have noticed that Sangria is running on a pretty ancient version of parboiled.

Upgrading also removes a fair amount of  warnings like:
```Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method unary_!```

For some reason scala-steward did not create a PR for it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,7 +27,7 @@ lazy val core = project
     testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-oF"),
     libraryDependencies ++= Seq(
       // AST Parser
-      "org.parboiled" %% "parboiled" % "2.1.8",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by agourlay** [comment]
Thanks to TravisCI now I now why the update has not been done 😭 

```
(sangria-core / update) sbt.librarymanagement.ResolveException: Error downloading org.parboiled:parboiled_2.11:2.2.1
```

yanns:
Do we know why we cannot download it?

agourlay:
Parboiled2 does not support Scala 2.11 since version `2.2.0`
https://github.com/sirthias/parboiled2/commit/672cb05134e7b8f73692a055b7d18be2d401070d

https://search.maven.org/search?q=g:org.parboiled%20AND%20a:parboiled_2.11

PR #533, opened 2020-12-07
Remove Scala 2.11 Support 🔥

nickhudkins:
2.11 targets JVM 7 which is old and stinky, this moves us into the future of Long Term support!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,9 @@ classes
 *.iml
 *.sc
 *.swp
+
+# Metals!
+.metals/
+.bloop/
+.vscode/
+metals.sbt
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,9 @@ classes
 *.iml
 *.sc
 *.swp
+
+# Metals!
+.metals/
+.bloop/
+.vscode/
+project/
```
yanns, line 16: I don't think we should ignore `project/`, right?

nickhudkins: Agreed :) a messy copy and paste

yanns (approved):
🥇

PR #548, opened 2020-12-16
Scalafmt

yanns:
@nickhudkins what do you think of adding auto-formatting?


zvuki:
Standardization/code formatting is great but I'm a bit worried that this can make merging/rebasing quite a mess if the folks who have improvements in their forks are willing to bring to main sangria (thinking Twitter and others...).

yanns:
@zvuki Thx for the feedback. I don't think it's a real issue.
- When looking into git history, it's easy to jump the "reformat" commit
- other forks just have to copy the scalafmt config file and reformat all files before proposing a PR.

nickhudkins:
I love this, and have been avoiding doing it because I didn't want to end up with a massive diff like this but it makes contributing EASIER and removes any silly nitpicks. Yes PLEASE! @yanns

PR #563, opened 2021-01-15
Support Args in ProjectedNames

nickhudkins:
After a little digging, thanks to @kirach's fork, I saw that the fix for the problem @yanns ran into https://github.com/sangria-graphql-org/sangria/pull/10#issuecomment-571471876 was a trivial fix.

I also updated the test cases to ensure queries with variables are a tested case.

Closing #468 in favor of this due to nasty rebasing from a long out of date branch :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,7 @@ import sbt.Keys.{
   scmInfo,
   startYear
 }
+import com.typesafe.tools.mima.core.{ProblemFilters, Problem}
 
 // sbt-github-actions needs configuration in `ThisBuild`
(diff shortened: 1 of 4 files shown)


nickhudkins:
Is it worth trying to make this binary compatible? It would add a pretty substantial amount of code just implementing `this, copy, apply, tupled` etc... and I am not sure that it is worth it, but happy to do so if that is of interest.

yanns (approved):
I don't think it's possible to make it binary compatible.
You can add an exception rule for this.

PR #571, opened 2021-01-29
federation WIP 1

yanns:


xsoufiane (approved):
I am 👍 for this, I was a bit sceptical because of the additional param and also about the fact that an unmarshaller is always linked to a marshaller. But again, I don't see an other way and we can always use the tests to ✅ so 💯

yanns:

PR #576, opened 2021-02-03
test: support map values as scalars

xsoufiane:


yanns:

PR #577, opened 2021-02-03
federation WIP 3

yanns:


modules/core/src/test/scala/sangria/federation/FederationSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+        AST,
+        AstSchemaBuilder.resolverBased[Any](
+          FieldResolver.map(
+            "Query" -> Map(
+              "entities" -> (ctx => ctx.withArgs(federationSupport.representationArg)(arg => arg))
+            )
+          ),
+          AdditionalTypes(federationSupport._Any.Type)
+        )
+      )
```
yanns, line 53: @xsoufiane I'm wondering if we could put those lines into `FederationSupport`. With schema extensions maybe?

xsoufiane: Great idea ❤️ I will spend some time trying to figure out if we can do it generically for all the other marshallers

modules/core/src/test/scala/sangria/federation/FederationSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+        override def isMapNode(node: Node): Boolean = default.isMapNode(node)
+        override def getMapValue(node: Node, key: String): Option[Node] =
+          default.getMapValue(node, key)
+        override def getMapKeys(node: Node): Traversable[String] = default.getMapKeys(node)
+        override def isListNode(node: Node): Boolean = default.isListNode(node)
+        override def getListValue(node: Node): Seq[Node] = default.getListValue(node)
+        override def isDefined(node: Node): Boolean = default.isDefined(node)
+        override def isEnumNode(node: Node): Boolean = default.isEnumNode(node)
+        override def isVariableNode(node: Node): Boolean = default.isVariableNode(node)
+        override def getScalaScalarValue(node: Node): Any =
```
xsoufiane, line 131: this also needs to change for test purposes 🤔

xsoufiane: In your case 🤔 NodeObject is not really a scala scalar

modules/core/src/test/scala/sangria/federation/FederationSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+        |    entities: "{\"__typename\":\"State\",\"id\":1}"
+        |  }
+        |}""".stripMargin)
+    }
+  }
+}
+
+object FederationSpec {
+
+  class FederationSupport[Node](default: InputUnmarshaller[Node]) {
```
yanns, line 82: The default `InputUnmarshaller` could be read from a `Schema` given as parameter.
The user API would be something like:
```
val standardSchema = ...
val federatedSchema = federation(standardSchema, ...)
```

xsoufiane: I taught about that too, it would be great if we can extract the marshaller from the schema and extend it the way we like. However, sangria separates them, and the marshaller is only defined at the executor level

modules/core/src/test/scala/sangria/federation/FederationSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+
+    trait NodeObject {
```
xsoufiane, line 84: I think we need some material than just the render method, I am working on refining the test because at the end we wanna be able to extract the information about an entity's keys from _Any

yanns:

PR #596, opened 2021-03-09
Introspection: Return interfaces implemented for interfaces

Tapped:
We should return interfaces implemented for interfaces even though it will be empty for now, since there is a bug in the GraphQL JS implementation which expects null or empty array: https://github.com/graphql/graphql-js/pull/2949

Currently Sangria doesn't handle interfaces implementing interfaces (RFC 373). Though I've made a quick take on it https://github.com/cognitedata/sangria/tree/add-rfc-373 but I've some failing tests which I've to figure out.


Tapped:
Nevermind, it's because of our custom Circe printer that we end up not compatible with the introspection logic in the GraphQL JS library.

PR #597, opened 2021-03-09
Add support for interfaces implementing interfaces

Tapped:
Add support for interfaces implementing interfaces

This is per the RFC 373 found here: https://github.com/graphql/graphql-spec/pull/373

TODO:
- [ ] Validate that there are no cycles
- [ ] Validate that transitively implemented interfaces (interfaces implemented by the interface that is being implemented) must also be defined on an implementing type or interface.
- [ ] Validate that the implementing interface must define each field that is specified by the implemented interface. 
- [ ] Add more tests

### DIFF
--- a/modules/core/src/main/scala/sangria/ast/QueryAst.scala
+++ b/modules/core/src/main/scala/sangria/ast/QueryAst.scala
@@ -451,6 +451,7 @@ case class ObjectTypeDefinition(
 
 case class InterfaceTypeDefinition(
     name: String,
+    interfaces: Vector[NamedType],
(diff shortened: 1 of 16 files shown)


yanns:
The binary compatibility checks are of course not happy with the changes. 😉 
You can disable them temporary the time to work on this PR.

Tapped:
Yes, I guess we can make it compatible by putting the interface field at the end of the interface with default = None.

yanns:
https://github.com/sangria-graphql/sangria/pull/981

PR #611, opened 2021-04-09
validate that different instances of InputObjectType with the same name are the same

yanns:
The validation only checks the field names. It cannot catch all cases, but I think it should be sufficient for 99% of the cases.

### DIFF
--- a/modules/core/src/main/scala/sangria/schema/Schema.scala
+++ b/modules/core/src/main/scala/sangria/schema/Schema.scala
@@ -1188,6 +1188,8 @@ case class Schema[Ctx, Val](
       (t1, t2) match {
         case (ot1: ObjectType[_, _], ot2: ObjectType[_, _]) =>
           sameSangriaType && (ot1.valClass == ot2.valClass)
+        case (ot1: InputObjectType[_], ot2: InputObjectType[_]) =>
+          sameSangriaType && (ot1.fieldsByName.keySet == ot2.fieldsByName.keySet)
         case _ => sameSangriaType
       }
     }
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/sangria/schema/Schema.scala
```diff
@@ -1232,12 +1238,10 @@ case class Schema[Ctx, Val](
               "You can find more info in the docs: http://sangria-graphql.org/learn/#circular-references-and-recursive-types")
         case t: Named if result contains t.name =>
           result.get(t.name) match {
-            case Some(found)
-                if !sameType(found._2, t) && t.isInstanceOf[ScalarAlias[_, _]] && found._2
-                  .isInstanceOf[ScalarType[_]] =>
-              result
             case Some(found) if !sameType(found._2, t) =>
-              typeConflict(t.name, found._2, t, parentInfo)
+              if (t.isInstanceOf[ScalarAlias[_, _]] && found._2.isInstanceOf[ScalarType[_]])
```
yanns, line 1242: I've extracted the `if` to avoid calling `sameType` several times.

xsoufiane (approved):
looks 💯, just a taught maybe you can also add a test to check the case of different field names (not just different classes)

nikola-mladenovic (approved):
💯

yanns:
> looks 💯, just a taught maybe you can also add a test to check the case of different field names (not just different classes)

Actually, we only test different field names as the class is not persisted with InputObjectType. I'll update the test name.

PR #612, opened 2021-04-14
support for configuring max concurrent batches in fetcher

nickhudkins:
added a new config param maxConcurrentBatches to configure
the max number of batched IDs that can be fetched concurrently by fetcher.
not specifying any value for this param will fallback to the existing
behaviour of running all batches concurrently.

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/deferred/FetcherBasedDeferredResolver.scala
+++ b/modules/core/src/main/scala/sangria/execution/deferred/FetcherBasedDeferredResolver.scala
@@ -182,6 +182,38 @@ class FetcherBasedDeferredResolver[-Ctx](
     }
   }
 
+  private def resolveConcurrentBatches(
+      ctx: FetcherContext[Ctx] @uncheckedVariance,
+      f: Fetcher[Ctx, Any, Any, Any] @uncheckedVariance,
+      nonCachedIds: Vector[Any]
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/sangria/execution/deferred/FetcherBasedDeferredResolver.scala
```diff
@@ -182,6 +182,34 @@ class FetcherBasedDeferredResolver[-Ctx](
...
+    nonCachedIds: Vector[Any]
+  )(implicit ec: ExecutionContext): Iterator[Future[(Vector[Any], Try[Seq[Any]])]] = {
+    val groupedIds: Iterator[Vector[Any]] = ctx.fetcher.config.maxBatchSizeConfig match {
+      case Some(size) ⇒ nonCachedIds.grouped(size)
+      case None ⇒ Iterator(nonCachedIds)
+    }
+
+    val groupedBatches: Iterator[TraversableOnce[Vector[Any]]] = ctx.fetcher.config.maxConcurrentBatchesConfig match {
+      case Some(size) ⇒ groupedIds.grouped(size)
+      case None ⇒ Iterator(groupedIds)
```
yanns, line 197: We are now using the `=>`. Can we change them?

nickhudkins: Replaced! I also updated `.scalafmt.conf` so no one has to ever think about it!

yanns (approved):
🥇

PR #620, opened 2021-04-24
Fixed obsoleted homepage URI

windymelt:
I fixed URI which provides homepage for Sangria from original one into github.io . Domain is already obsoleted.

Thanks.


yanns (approved):
💯

PR #674, opened 2021-06-04
Minor performance optimization to ValueCoercionHelper.valuePosition

agourlay:
This function shows up while profiling mostly due to the creation of the `View` which is not strictly necessary.

The proposed implementation only relies on `collectFirst` which should be significantly faster than `view.collect.headOption`


sh0hei (approved):

sh0hei:
@agourlay Thanks a lot !

PR #716, opened 2021-09-03
document some of the AST & parser packages

performantdata:
Also:
- adds type annotations to public members, to address IDE warnings
- adds `override` modifiers
- moves some types from the huge QueryAst.scala file to their own files


yanns (approved):
💯

performantdata:
> Could you add a rule

Done. I also added some more documentation changes.

yanns:
> > Could you add a rule
> 
> Done. I also added some more documentation changes.

Thanks. The build is still failing though. Can you have a look?

performantdata:
Yeah, I'm still looking at it. It's because `ListMap.keys` returns an `immutable.Iterable`, while `Map.keys` doesn't.

performantdata:
How do you plan to deal with these binary incompatibilities @yanns ? Are you going to bump the major version, or just hope that they don't affect anyone? I guess it's all the same to me for now, just another MiMa exception.

yanns:
> How do you plan to deal with these binary incompatibilities @yanns ? Are you going to bump the major version, or just hope that they don't affect anyone? I guess it's all the same to me for now, just another MiMa exception.

If they are minimal, I don't think a major version is necessary for that.

yanns (approved):
💯

PR #724, opened 2021-09-15
document and clean the AST code

performantdata:
Documentation and clean-up of IDE warnings related to my investigation into separating the AST package into its own module.

### DIFF
--- a/modules/core/src/main/scala/sangria/ast/QueryAst.scala
+++ b/modules/core/src/main/scala/sangria/ast/QueryAst.scala
@@ -204,7 +204,13 @@ sealed trait WithTrailingComments {
   def trailingComments: Vector[Comment]
 }
 
+/** A GraphQL AST node that contains [[Selection selections]].
+  *
+  * Most typically, this is a [[Field field]] that is of a composite type.
+  */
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/sangria/ast/QueryAst.scala
```diff
@@ -498,6 +514,8 @@ case class Comment(text: String, location: Option[AstLocation] = None) extends A
 
 // Schema Definition
 
+/** @group typesystem
```
yanns, line 517: What is this `@group`? is if the scaladoc?

yanns (approved):
💯

modules/core/src/main/scala/sangria/ast/QueryAst.scala
performantdata: https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#member-grouping-tags

performantdata: You can already see an example [here](https://www.javadoc.io/doc/org.sangria-graphql/sangria_2.13/latest/sangria/ast/index.html).

PR #751, opened 2021-10-15
Do not force QueryValidator to use List

bjing:
Do not force the use of `List` type for rules when creating a `QueryValidator`. 

List is not a very efficient type and forcing the clients to create rules in `List` has no obvious benefit because Sangeria is not using any functionality from an actual `List` type.

This gives the users the flexibility to use the types they need, if they need `List` or `Vector` or anything else, this will allow them to do it.

Also, for the `def validateUsingRules()` function, since visitors parameter only gets iterated once, there's no need to mandate any collection type other than an `Iterable`.

### DIFF
--- a/modules/core/src/main/scala/sangria/validation/QueryValidator.scala
+++ b/modules/core/src/main/scala/sangria/validation/QueryValidator.scala
@@ -16,7 +16,7 @@ trait QueryValidator {
 }
 
 object QueryValidator {
-  val allRules: List[ValidationRule] = List(
+  val allRules: Seq[ValidationRule] = Seq(
     new ValuesOfCorrectType,
     new ExecutableDefinitions,
     new FieldsOnCorrectType,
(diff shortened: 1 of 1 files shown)


yanns:
Sorry, I somehow missed that PR.
Do you have examples when this optimization is relevant?

The list of rules is normally created once with the schema for the whole application's life. Given that, I am not sure if this optimization brings much value.
But maybe you're using the library another way?

bjing:
Hi @yanns thanks for getting back to me. I no longer work at the place where this change is needed, and I can no longer provide an example where this causes an issue, it was indeed an issue for us back then. 

However, from a design point of view, I think it's always nice to use more of a base type (or interface) instead of something like a `List`, especially when you're not using any features from a `List`. This would give the users more flexibility in terms of what they can do.

Anyway, since I no longer work with Sangria, feel free to disregard this PR.

yanns:
I see your point.
But this change is not backwards compatible. I need to check if the reasons are good enough.
Let me think about it.

Issue #466, opened 2019-11-08
Split sangria parse and AST into separate module


Slakah:
Related to https://github.com/sangria-graphql/sangria/issues/461 . I might be looking to write a sangria client, and moving the parsing out would be very helpful.

Hopefully I'll find time to start working on this over the next couple of weeks.

Issue #461, opened 2019-09-26
Split core sangria into modules


paulpdaniels:
Currently the sangria core package includes everything needed to start working with sangria on the server side. However, there are cases where this seems to be overkill like with the drunk client. I propose splitting the core module into several sub modules that will continue to live here. For ease of use we’d continue to export the core module which would contain all modules but would allow consumers to pick more specific modules depending on their needs. My thought is:

- sangria (aggregate)
- sangria-execution (Executor/reducers)
- sangria-derived (macros and derivation)
- sangria-ast (The schema ast)
- sangria-parse (just the ast parsing)

The advantage here is that the parse module could be exchanged with a different parser, clients only need ship with a hard dependency on the ast. And lib developers can explore different execution models while still relying on solid foundations.

performantdata:
There is now [a `sangria-ast` library](https://search.maven.org/search?q=g:org.sangria-graphql%20a:sangria-ast_2.13).

performantdata:
There is now [a `sangra-parser` library](https://search.maven.org/search?q=g:org.sangria-graphql%20a:sangria-parser_2.13).

yanns:
This is now a [sangria-derivation library](https://search.maven.org/artifact/org.sangria-graphql/sangria-derivation_2.13)
I don't think we will introduce more modules.

PR #777, opened 2021-10-30
AST separation

performantdata:
This involves separating the AST code into its own published module. Deprecation of what we throw away or change in the API should happen in a new 2.x branch, if desired.

This PR will help solve #461 and #466.


performantdata:
I based this branch on #724.

modules/ast/src/main/scala/sangria/ast/SourceMapper.scala
```diff
@@ -1,4 +1,4 @@
-package sangria.parser
+package sangria.ast
```
yanns, line 1: Maybe we can revert this change? This would result in much less changes.

modules/benchmarks/src/main/scala/sangria/benchmarks/OverlappingFieldsCanBeMergedBenchmark.scala
```diff
@@ -119,7 +116,7 @@ class OverlappingFieldsCanBeMergedBenchmark {
                |     xingId {
              """.stripMargin)
 
-    for (i <- 1 to size)
+    for (_ <- 1 to size)
```
yanns, line 119: It'd help the review if you can keep the diff as minimal as possible.

yanns:
I'll try to use a snapshot later to check how important the breaking changes are.

build.sbt
```diff
@@ -178,9 +308,23 @@ lazy val root = project
...
   .disablePlugins(MimaPlugin)
 
+lazy val ast = project.in(file("modules/ast"))
+  .withId("sangria-ast")
+  .settings(scalacSettings ++ shellSettings)
+  .settings(
+    name := "sangria-ast",
+    description := "Scala GraphQL AST representation",
+
+    //FIXME This really shouldn't depend on any other libraries.
```
yanns, line 318: Avoid FIXME in the code that are never fixed ;)
Better adding an issue and linking the issue here.

modules/ast/src/main/scala/sangria/ast/SourceMapper.scala
performantdata: The `SourceMapper` belongs with the `ast` library. Well, not all of it. The `render` methods don't belong there. But the identification of the source and the content of the source belong there.

But yes, this is the main change. It was intentional.

build.sbt
performantdata: Not sure what you mean. A "FIXME" will show up in the IDE list of problems. And the solution to this one is removing the `render` methods, which…will get done eventually? next, even? That's part of the discussion.

There are a lot of issues that don't have GitHub issues tracking them. And I don't even know whether they're issues, because it's hard to have that discussion with you guys, over Gitter.

modules/ast/src/main/scala/sangria/ast/SourceMapper.scala
performantdata: My _belief_ is that it isn't disruptive. I doubt that users create their own `SourceMapper`s, or use the `SourceMapper`s directly. The only reason that i could see for doing so would be internationalization of the error messages, and then there are a lot more such messages in the `Violation`s.

build.sbt
yanns: In my experience, FIXME in the code stay forever and are never tackled. That's why I try to avoid them now.
But having that as a github issue will not ensure that it will be tackle. ;) But at least it's more visible to eventual contributors.

performantdata: You realize of course that you have six-year-old GitHub issues in the project. 😏 And the build spits out 100+ warnings, with more warnings from IntelliJ. To me, a "FIXME" or "TODO" is just another tool in the drawer. It's useful when the issue doesn't warrant a full-blown GitHub issue. The IDE tracks them, in the Problems tab.

This one in particular I expect to be addressed on the way to separating out the localized string from the Sangria core code. _That_ warrants a GH issue, IMO.

yanns (commented):
I've opened a PR with some suggestions: https://github.com/performantdata/sangria/pull/3

modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala
```diff
@@ -485,6 +488,13 @@ object QueryRenderer {
   def render(node: AstNode, config: QueryRendererConfig = Pretty, indentLevel: Int = 0): String =
     renderNode(node, config, Indent(config, indentLevel, indentLevel))
 
+  def renderPretty(node: AstNode): String = node match {
+    case _: Value => render(node, PrettyInput)
+    case _ => render(node, Pretty)
+  }
+
+  def renderCompact(node: AstNode): String = render(node, Compact)
```
yanns, line 496: This would be something to backport to the 2.x version before, deprecating the `.renderXXX` methods on the AST. This would be a migration path to smoothly transition to this,

performantdata: Yep. I have [ideas of what to deprecate](https://github.com/performantdata/sangria/discussions/1#discussioncomment-1573890). Just need to see what gets committed.

performantdata:
https://github.com/performantdata/sangria/pull/3 is cherry-picked.

modules/core/src/main/scala/sangria/macros/AstLiftable.scala
```diff
@@ -142,12 +142,12 @@ trait AstLiftable {
   }
 
   implicit def liftDocument: Liftable[Document] = Liftable { case doc @ Document(d, c, p, _) =>
-    q"_root_.sangria.ast.Document($d, $c, $p, _root_.scala.Some(new _root_.sangria.ast.DefaultSourceMapper(${doc.sourceMapper.get.id}, _root_.org.parboiled2.ParserInput(${doc.source.get}))))"
+    q"_root_.sangria.ast.Document($d, $c, $p, _root_.scala.Some(new _root_.sangria.parser.DefaultSourceMapper(${doc.sourceMapper.get.id}, _root_.org.parboiled2.ParserInput(${doc.source.get}))))"
```
yanns: What I forgot is that we need to use `SourceMapperInput` here also.
I'm split between having a helper method converting `org.parboiled2.ParserInput` to `SourceMapperInput`.
Or making `SourceMapperInput` using implicit to find the right instance so that we "only" need to import the right package.

yanns: I see now that you did that here: https://github.com/sangria-graphql/sangria/pull/777/commits/ed755de5dfe3e4d286a18a4b921b965a9b55d0c9
👍

modules/core/src/main/scala/sangria/package.scala
```diff
@@ -0,0 +1,15 @@
+/** Sangria GraphQL server library.
+  *
+  * Sangria is a library that provides parsing, validation, execution and other services for GraphQL
+  * queries.
+  *
+  * It typically requires other libraries to build a complete GraphQL service—including, perhaps,
+  * ones that provide a service interface (e.g.,
+  * [[https://doc.akka.io/docs/akka-http/current/index.html Akka HTTP]]) and a database interface
+  * (e.g., [[https://scala-slick.org/ Slick]])—as well as custom code to bridge the gap between data
+  * representations that are natural to the database vs. to the GraphQL schema.
```
yanns, line 10: My experience is that new users read the website and not the scaladoc.
So I'd prefer not linking to any other libraries here, keeping the sources neutral.

```suggestion
  * It typically requires other libraries to build a complete GraphQL service
  * —including, perhaps, ones that provide a HTTP service interface and a database interface—
  * as well as custom code to bridge the gap between data representations that are natural
  * to the database vs. to the GraphQL schema.
```

modules/core/src/main/scala/sangria/parser/package.scala
```diff
@@ -0,0 +1,7 @@
+package sangria
+
+package object parser {
+
+  @deprecated("please use `sangria.ast.SourceMapper`", "3.0.0")
+  type SourceMapper = sangria.ast.SourceMapper
```
yanns, line 6: Should we add a small test to make sure we provide this backwards compatibility?
During a migration from 2.x to 3.x, this is very important to avoid compilation errors.

yanns (commented):
This is looking better & better! 👍

modules/core/src/main/scala/sangria/parser/package.scala
performantdata: What are you thinking? What would the test be?

yanns: Very simple, just using `sangria.parser.SourceMapper`.
```
val mapper: sangria.parser.SourceMapper = new DefaultSourceMapper...
succeeded
```

modules/core/src/main/scala/sangria/package.scala
performantdata: I also recognize that new users use the website docs first. I don't see how that's relevant. Yes, this scaladoc is for the users who are digging for a deeper understanding.

I felt that pointing to a specific example of "an HTTP service interface" was clearer than using a term ("HTTP service interface") that may be ambiguous to the reader. I picked the ones that I thought most users would know. My text first read like yours, without the "(e.g., …)"s.

yanns: You can change the text how you like.
I'd just like that we don't mention any other libraries / projects to stay neutral.

modules/core/src/main/scala/sangria/parser/package.scala
performantdata: Oh, like:
```scala
doesThisCompile {
  val mapper: sangria.parser.SourceMapper = new DefaultSourceMapper...
}
```
I haven't built a test like that yet.

I'm busy most of today, but can maybe get to it later.

performantdata:
:shipit:

performantdata:
How/when does the CHANGELOG get updated?

performantdata:
This should be added to the milestone. @yanns

modules/core/src/main/scala/sangria/package.scala
yanns: Can you please change that?

yanns (changes requested):
Great work!
Can you please remove the references to other libraries? Thanks!

yanns:
> How/when does the CHANGELOG get updated?

manually, when we can.

modules/core/src/main/scala/sangria/package.scala
performantdata: > You can change the text how you like.

I thought you were giving me the choice.

yanns (approved):
💯

modules/core/src/main/scala/sangria/package.scala
yanns: You can change the text how you like *as long as* we don't mention any other libraries / projects to stay neutral. 😉

yanns:
This change breaks some projects: https://github.com/sangria-graphql/sangria/issues/795

PR #781, opened 2021-11-03
backport ExecutionPath.cacheKeyReversed

yanns:
From https://github.com/sangria-graphql/sangria/pull/775
and https://github.com/sangria-graphql/sangria/pull/774

By using this in sangria-slow-logs, we will be able to generate much less objects.

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/ExecutionPath.scala
+++ b/modules/core/src/main/scala/sangria/execution/ExecutionPath.scala
@@ -29,6 +29,8 @@ case class ExecutionPath private (path: Vector[Any], cacheKeyPath: ExecutionPath
   })
 
   def cacheKey: ExecutionPath.PathCacheKey = cacheKeyPath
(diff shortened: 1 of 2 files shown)


yanns:
The code I'd like to optimize is the following one (from sangria-slow-logs):
```
    val path = even(ctx.path.cacheKey)

    queryVal.update(path, ctx.parentType.name, ctx.field.name, success, fieldVal, System.nanoTime())
  }

  def even[T](v: Vector[T]): Vector[T] = v.iterator.zipWithIndex
    .filter(_._2 % 2 == 0)
    .map(_._1)
    .toVector
```

This new "API" does not really help here.

yanns:
Actually the reverse iterator can be helpful:
```
    val path = even(ctx.path.cacheKey.reverseIterator)

  private def even[T](v: Iterator[T]): Vector[T] = {
    val l = new _root_.scala.collection.mutable.ListBuffer[T]()
    var even = false
    while (v.hasNext) {
      val e = v.next()
      if (even) {
        l.prepend(e)
      }
      even = !even
    }
    l.toVector
  }
```

PR #784, opened 2021-11-07
deprecate & backport for AST separation

performantdata:
Follow-up to #777.

### DIFF
--- a/modules/core/src/main/scala/sangria/ast/QueryAst.scala
+++ b/modules/core/src/main/scala/sangria/ast/QueryAst.scala
@@ -1,12 +1,12 @@
 package sangria.ast
 
 import sangria.execution.InputDocumentMaterializer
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/sangria/ast/QueryAst.scala
```diff
@@ -767,7 +829,15 @@ sealed trait ObjectLikeTypeExtensionDefinition extends TypeExtensionDefinition {
   def fields: Vector[FieldDefinition]
 }
 
+@deprecated(
+  "Renamed to sangria.schema.AstNodeTransformer in 3.0.",
```
yanns, line 833:
```suggestion
  "Removed in 3.0. Use sangria.schema.AstNodeTransformer instead.",
```

yanns (approved):
🥇

PR #790, opened 2021-11-10
remove `DeliveryScheme`

### LINKED ISSUE #783: remove `DeliveryScheme`
I'm raising this issue now, since we're talking about moving the `parser` package into its own library, and `DeliveryScheme` is part of that package. But the presence of this class throughout the Sangria code has been an annoyance.

Apparently [Oleg got the idea from the Parboiled2 project](https://github.com/sangria-graphql/sangria/commit/b60d2bb6311bea353f108d647a4e275c7ff7f5bb). It was [one of his earliest commits](https://github.com/sangria-graphql/sangria/commits/b60d2bb6311bea353f108d647a4e275c7ff7f5bb), and I suspect that it was a premature design choice. [The original idea in Parboiled2](https://github.com/sirthias/parboiled2/blob/b504cd883d37a50e1938ab2d3be0e0d501316251/README.rst#alternative-deliveryschemes) seems to be of dubious value itself.

So my suggestion is that, in the name of simplifying the code, that we return a `Try` where needed. This can easily be turned into an `Either` or a thrown exception.

performantdata:
Uses are replaced with `Try`. Fixes #783.


modules/core/src/test/scala/sangria/execution/InputDocumentMaterializerSpec.scala
```diff
@@ -192,13 +192,14 @@ class InputDocumentMaterializerSpec extends AnyWordSpec with Matchers with Strin
...
           }
         """)
+        .get
 
       val vars = scalaInput(
         Map(
           "comm" -> "from variable"
         ))
 
-      InputDocumentMaterializer.to(document, ArticleType, vars) should be(
```
performantdata, line 201: This is the sort of implicits magic that I don't like to see. `to()` throws an exception? instead of returning a `Try`? Oh, that's because of the import 200 lines up. One reason that I don't like `DeliveryScheme`.

modules/core/src/test/scala/sangria/util/CatsSupport.scala
```diff
@@ -215,26 +215,23 @@ object CatsScenarioExecutor extends FutureResultSupport {
...
+    case Execute(validate, value, vars, op) =>
       val validator = if (validate) QueryValidator.default else QueryValidator.empty
 
       ExecutionResult(
         Try(
           Executor
             .execute(
               given.schema,
-              QueryParser.parse(given.query),
+              QueryParser.parse(given.query).get,
```
yanns, line 234: Instead of calling `.get`, we can `map` on it?
```
QueryParser.parse(given.query).map(ast => Executor.execute(given.schema, ast,...))
```

performantdata: One could `map` on it, but that would change the behavior of the code. The `import sangria.parser.DeliveryScheme.Throw` above it says that `QueryParser.parse` should throw an exception on failure.

yanns: This exception would be catch by the `Try` above.

performantdata: I see. You're proposing to replace lines 230-240 in the new code. You'd need the `.await` as well. Sure, one could do that. My goal wasn't to analyze the code to make it aesthetic, just trying to get the PR done.

yanns (approved):

PR #791, opened 2021-11-12
deprecate `DeliveryScheme`

### LINKED ISSUE #783: remove `DeliveryScheme`

performantdata:
Fixes #783.

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/InputDocumentMaterializer.scala
+++ b/modules/core/src/main/scala/sangria/execution/InputDocumentMaterializer.scala
@@ -1,7 +1,7 @@
 package sangria.execution
 
 import sangria.ast.{AstVisitor, InputDocument, VariableDefinition}
-import sangria.ast
+import sangria.{ast, since2_1_7}
 import sangria.marshalling.{FromInput, InputUnmarshaller}
 import sangria.parser.DeliveryScheme
 import sangria.renderer.SchemaRenderer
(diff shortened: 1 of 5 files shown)


performantdata:
The deprecations are overly aggressive, coming up when there isn't any explicit use of the implicits. For instance, this comes from a new-style usage:
```
[warn] /home/michael/git/sangria/modules/core/src/test/scala/sangria/execution/ExecutorSpec.scala:1019:43: method Try in object DeliveryScheme is deprecated (since Sangria 2.1.7): Removed in 3.0. A Try result will be returned by default.
[warn]       val Success(doc) = QueryParser.parse("{eager, future}")
[warn]                                           ^
```

Better safe than sorry? Anyone who has warning set to fatal will need to `@nowarn`.

performantdata:
@yanns This too for #783.

yanns:
it's too aggressive.
Existing code not importing `DeliveryScheme` should not have any warnings.

performantdata:
> Existing code not importing `DeliveryScheme` should not have any warnings.

I whittled it back to that standard.

yanns (approved):
👍

PR #808, opened 2021-11-23
deprecate renderers in schema/introspection code

### LINKED ISSUE #806: remove the renderers from the schema/introspection code
There are calls to `SchemaRenderer` and `QueryRenderer` buried in the schema and introspection classes. They seem non-essential and create an otherwise unneeded dependency from the schema/introspection representations to the renderers.

This is done with an eye towards separating out the schema/introspection representations, and maybe the renderers, from the `sangria-core` code.

performantdata:
Fixes #806.


performantdata:
#806 is replaced by #823.

PR #825, opened 2021-12-08
split the schema representation classes into their own module

### LINKED ISSUE #823: split the schema representation classes into their own module
This relates to #461.

It appears to me that the schema representation classes have no need to depend on the rest of the https://github.com/sangria-graphql/sangria classes except the `ast` module. I think that I can split off the schema representation classes, mostly `Schema.scala`, `Context`, `Violation` and some others. I think there should be modules that parallel `ast` and `parser`: one that creates the representation, and the representation itself. I currently expect the module dependencies to look like this:

Sangria components

performantdata:
Fixes #823.


yanns:
When I look at [all impacts on existing users](https://github.com/sangria-graphql/sangria/pull/824), I'm tending not to do this modularization. The trade-of is tending towards too much pain for users, and a library that is less easy to use.

performantdata:
> When I look at [all impacts on existing users](https://github.com/sangria-graphql/sangria/pull/824), I'm tending not to do this modularization. The trade-of is tending towards too much pain for users, and a library that is less easy to use.

Not sure where your concerns lie, but one approach that I haven't incorporated into mitigating changes is to "pimp" the libraries with extension methods. Meaning that, there are many instances where I've removed a method because it was using code that is, after modularization, no longer part of its library or of any required library. In those cases, the methods could be added back as extension methods in a "higher-level" library. The only change then would be adding an import. Of course, there's no way to warn for that via deprecation.

yanns:
> > When I look at [all impacts on existing users](https://github.com/sangria-graphql/sangria/pull/824), I'm tending not to do this modularization. The trade-of is tending towards too much pain for users, and a library that is less easy to use.
> 
> Not sure where your concerns lie, but one approach that I haven't incorporated into mitigating changes is to "pimp" the libraries with extension methods. Meaning that, there are many instances where I've removed a method because it was using code that is, after modularization, no longer part of its library or of any required library. In those cases, the methods could be added back as extension methods in a "higher-level" library. The only change then would be adding an import. ~Of course, there's no way to warn for that via deprecation.~

I understand. It's still a pain for users.
It's a trade-of here, with no right or wrong, and I think it's better in that case to favor the user experience over nice modularization.

PR #866, opened 2022-04-06
Introspection complexity for TypeNameMetaField

xsoufiane:
please review 🙏 

### Motivation

Currently sangria has a default complexity of 1 for each field, even for introspection field. In this PR, I am making it possible to have a different complexity value for the introspection __typename field, by adding a [config file]( modules/core/src/main/resources/reference.conf) with the relative settings.

### Notes for reviewer

- The PR doesn't break the default complexity setting for __typename. I ensured this by adding [reference.conf](modules/core/src/main/resources/reference.conf) which defaults to 1. Applications can change the value by adding a different setting in application.conf (example is in the unit tests).

- To be able to use different configurations in the tests, I needed to make TypeNameMetaField a function: https://github.com/sangria-graphql/sangria/blob/04736fe703e8e69dd52be7b381f4c76840211357/modules/core/src/main/scala/sangria/introspection/package.scala#L428 
this will probably add some time overhead but it's the only simple way that I could think of to do this without changing the overall structure. If there are any concerns about this please let me know.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,7 +100,10 @@ lazy val core = project
       "eu.timepit" %% "refined" % "0.9.28" % Test,
       // CATs
       "net.jcazevedo" %% "moultingyaml" % "0.4.2" % Test,
-      "io.github.classgraph" % "classgraph" % "4.8.143" % Test
+      "io.github.classgraph" % "classgraph" % "4.8.143" % Test,
+      // Config
+      "com.typesafe" % "config" % "1.4.2",
+      "com.iheart" %% "ficus" % "1.5.2"
     ),
     apiURL := {
       val ver = CrossVersion.binaryScalaVersion(scalaVersion.value)
@@ -202,3 +205,5 @@ lazy val shellSettings = Seq(
 lazy val noPublishSettings = Seq(
   publish / skip := true
 )
+
+logLevel := Level.Error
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -100,7 +100,10 @@ lazy val core = project
       "eu.timepit" %% "refined" % "0.9.28" % Test,
       // CATs
       "net.jcazevedo" %% "moultingyaml" % "0.4.2" % Test,
-      "io.github.classgraph" % "classgraph" % "4.8.143" % Test
+      "io.github.classgraph" % "classgraph" % "4.8.143" % Test,
+      // Config
+      "com.typesafe" % "config" % "1.4.2",
+      "com.iheart" %% "ficus" % "1.5.2"
```
yanns, line 106: Can we try to avoid adding an external dependency?

xsoufiane: I don't see a way, not using these libs will force us to write our own configuration library 🤔 or change the structure of sangria so that users supply a config object when using it but this is the complex way

yanns: OK I'll try on myself then.

xsoufiane: gonna close this PR because there is another simple way to achieve this by overriding the MeasureComplexity class reduceField function at the application level https://github.com/sangria-graphql/sangria/blob/09f744992477f2310e508a6dc7786353a6d51699/modules/core/src/main/scala/sangria/execution/QueryReducer.scala#L77 Thanks @yanns  for the super idea.

PR #881, opened 2022-06-20
Check a sangria type declared multiple times to have the same fields

AL333Z:
Hi 👋 ,

Last week I spotted an unexpected behaviour while cleaning our schemas.. and it turned out the main cause is that Sangria won't raise a violation if there're two `ObjectType` with the same `name` and `valClass`, but different field set.

This PR will extend the current validation by also checking the fields set.
I added a new test case to cover the behaviour.

Not sure about the impact of this little change.. is ok to change that straight away or would it better to make this check configurable and opt-in?


yanns:
This change can be tried out in https://github.com/sangria-graphql/sangria/releases/tag/v3.1.0-RC1

PR #883, opened 2022-06-23
Migrate the core module and tests to scala 3

jchyb:
<s>All of the tests but one pass (the failing one is `deserialize automatically with spray-json JsonFormat (optional list with optional values and default)` in FromInputSpec - I have not yet figured out why that is.</s> All of the tests should pass now.
<s>
For the code in the PR to work some dependencies need to be released for scala 3 and have their version updated. I commented near those dependencies with "TODO needs release", those releases should contain the following PRs to work:
* [macro-visit](https://github.com/sangria-graphql/macro-visit/pull/135)
* [sangria-marshalling-api](https://github.com/sangria-graphql/sangria-marshalling-api/pull/132)
</s>

While some changes to the codebase may seem questionable, they are the result of changed type inference, and, in a few cases, bugs:
* [Necessary explicit `Argument` type parameters](https://github.com/lampepfl/dotty/issues/15365)
* [Some default parameters not working](https://github.com/lampepfl/dotty/issues/15287)

I imagine once those become fixed, the various workarounds used could be removed.

I still have to complete the sangria-derivation module to finish the migration, which may take some time. However the basic functionality of defining graphql code and executing queries should be already working.


yanns:
The following releases can be used now:
- https://github.com/sangria-graphql/macro-visit/releases/tag/v0.1.3
- https://github.com/sangria-graphql/sangria-marshalling-api/releases/tag/v1.0.8

jchyb:
I finally managed to fix all of the tests, with the one mentioned above being more important than I initially though. sangria-derivation module is also finished and working (with a small exception), but I think I should make a separate PR for that.

yanns:
There are 2 scalac options that we should deactivate for scala3:
```
[warn] bad option '-Xlint:-missing-interpolator,_' was ignored
[warn] bad option '-target:jvm-1.8' was ignored
```
Example on how to deal with that: https://github.com/sangria-graphql/sangria-spray-json/blob/main/build.sbt#L27

modules/core/src/main/scala/sangria/schema/AstSchemaResolver.scala
```diff
@@ -223,7 +225,7 @@ case class GenericDynamicDirectiveResolver[T, A](
     directiveName: String,
     locations: Set[DirectiveLocation.Value] = Set.empty,
     resolve: GenericDynamicDirectiveContext[A] => Option[T])(implicit
-    val marshaller: ResultMarshallerForType[T])
+    val marshaller: ResultMarshallerForType[A])
```
yanns, line 228: Was this a bug?

yanns (commented):
Thanks a lot for all this effort!! ❤️

yanns:
The `mimaReportBinaryIssues` should not be triggered on scala 3.
This can be configured like in https://github.com/sangria-graphql/sangria-spray-json/blob/main/build.sbt#L6-L12

modules/core/src/main/scala-3/sangria/macros/ToExprGivens.scala
```diff
@@ -0,0 +1,414 @@
...
+      case DirectiveDefinition(n, a, l, desc, c, p) =>
+        '{
+          sangria.ast.DirectiveDefinition(
+            ${ Expr(n) },
+            ${ Expr(a) },
+            ${ Expr(l) },
+            ${ Expr(desc) },
+            ${ Expr(c) },
+            ${ Expr(p) })
+        }
```
yanns, line 137: One field was added in the `DirectiveDefinition` in https://github.com/sangria-graphql/sangria/pull/899 that should be backported here to fix the compilation error:
```
[error] 128 |      case DirectiveDefinition(n, a, l, desc, c, p) =>
[error]     |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |Wrong number of argument patterns for sangria.ast.DirectiveDefinition; expected: (String, Vector[sangria.ast.InputValueDefinition], Vector[sangria.ast.DirectiveLocation], Option[sangria.ast.StringValue], Boolean, Vector[sangria.ast.Comment], Option[sangria.ast.AstLocation])
```

```suggestion
      case DirectiveDefinition(n, a, l, desc, r, c, p) =>
        '{
          sangria.ast.DirectiveDefinition(
            ${ Expr(n) },
            ${ Expr(a) },
            ${ Expr(l) },
            ${ Expr(desc) },
            ${ Expr(r) },
            ${ Expr(c) },
            ${ Expr(p) })
        }
```

modules/core/src/main/scala/sangria/schema/AstSchemaResolver.scala
jchyb: Yes, without this change I would get an error later on:
```
[error] -- [E007] Type Mismatch Error: /Users/jchyb/Documents/workspace/sangria/modules/core/src/main/scala/sangria/schema/ResolverBasedAstSchemaBuilder.scala:681:52 
[error] 681 |                          createDynamicArgs(astDir)(marshaller)))
[error]     |                                                    ^^^^^^^^^^
[error]     |Found:    (marshaller : sangria.marshalling.ResultMarshallerForType[T])
[error]     |Required: sangria.marshalling.ResultMarshallerForType[A$2]
[error]     |
[error]     |where:    A$2 is a type in an anonymous function in method resolveDirectives with bounds 
[error]     |
[error]     | longer explanation available when compiling with `-explain`
```

modules/core/src/main/scala/sangria/schema/Context.scala
```diff
@@ -164,7 +164,7 @@ object UpdateCtx {
 }
 
 private[sangria] case class SubscriptionValue[Ctx, Val, S[_]](
-    source: Val,
+    source: S[Any],
```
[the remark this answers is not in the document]
jchyb: Yes, leading to multiple compilation errors:
```
[error] -- [E007] Type Mismatch Error: /Users/jchyb/Documents/workspace/sangria/modules/core/src/main/scala/sangria/execution/Resolver.scala:268:64 
[error] 268 |            val s = svalue.stream.mapFuture[Any, Result](svalue.source) { action =>
[error]     |                                                         ^^^^^^^^^^^^^
[error]     |Found:    (svalue.source : Val$1)
[error]     |Required: S$2[Any]
[error]     |
[error]     |where:    S$2 is a type in an anonymous function in method resolveSubs with bounds <: [_] =>> Any
[error]     |
[error]     | longer explanation available when compiling with `-explain`
[error] -- [E007] Type Mismatch Error: /Users/jchyb/Documents/workspace/sangria/modules/core/src/main/scala/sangria/schema/Schema.scala:657:35 
[error] 657 |          resolve(ctx).asInstanceOf[stream.StreamSource[Any]],
[error]     |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |Found:    stream.StreamSource[Any]
[error]     |Required: StreamSource²
[error]     |
[error]     |where:    StreamSource  is a type in trait SubscriptionStreamLike with bounds <: [_] =>> Any
[error]     |          StreamSource² is a type in method subs
[error]     |
[error]     | longer explanation available when compiling with `-explain`
```

jchyb:
I had some trouble with disabling sangria-derivation compilation in Scala 3 (despite not having sources to compare to, mima plugin would still try to compile sangria-derivation), so for now I've removed Scala 3 from sbt-github-actions and set up simple separate GHA manually for scala 3 (via the ci-scala3.yml file). Later when PR'ing sangria-derivation I will remove that file and reenable the previous sbt-github-actions for scala 3

yanns:
> I had some trouble with disabling sangria-derivation compilation in Scala 3 (despite not having sources to compare to, mima plugin would still try to compile sangria-derivation), so for now I've removed Scala 3 from sbt-github-actions and set up simple separate GHA manually for scala 3 (via the ci-scala3.yml file). Later when PR'ing sangria-derivation I will remove that file and reenable the previous sbt-github-actions for scala 3

That's a reasonable approach for the time being! 👍

modules/core/src/main/scala/sangria/schema/Schema.scala
```diff
@@ -70,7 +70,7 @@ sealed trait AbstractType extends Type with Named {
       .flatMap(_.find(_.isInstanceOf(value)).asInstanceOf[Option[ObjectType[Ctx, _]]])
 }
 
-sealed trait MappedAbstractType[T] extends Type with AbstractType with OutputType[T] {
+sealed trait MappedAbstractType[T] extends Type with AbstractType {
   def contraMap(value: T): Any
 }
```
jchyb, line 75: I see that this change causes binary compatibility errors (for scala 2). I removed OutputType[T] as it would cause compiler conflicts in scala 3 (there is another place where an anonymous class is created with both OutputType[T] from MappedAbstractType and OutputType[Any] from UnionType). Removing it here did not cause any problems elsewhere, so I did not think much of it. Since the this is a sealed trait inside a file full of them, making a version specific versions of the file seems impractical and heavy on the maintenance work... Outside of breaking binary compatibility the only solution I can think of is something akin to https://github.com/akka/akka-http/pull/4113, where they have an sbt plugin that filters some implementations out of a single file depending on the scala version used

yanns: You can also add some `mimaBinaryIssueFilters`.

yanns (approved):
💯 🥇

yanns:
For me, the PR is looking ready to merge to that we can iterate on that. Or do you still have something to change on it?

yanns:
Can you check is the publishing part would only publish the modules that are ready for scala 3?
`sangria-derivation` should not be published for scala 3 for example.

jchyb:
I think we can merge, anything else that I have not committed yet concerns sangria-derivation module, which is separate. About the publishing, the original GitHub action yml now does not have any mention of Scala 3 (as I purposefully removed it for now), and the new one I created does not have the publish action. I was planning to reenable the publish action via the original yaml when I PR sangria-derivation (which is ready, outside of some small issues which I am not sure I can fix)

yanns:
One more advantage of a separated GH action! Cool!

PR #898, opened 2022-08-03
mandatory fields with default values are considered optional

yanns:
http://spec.graphql.org/draft/#sec-Input-Object-Required-Fields


clemniem (approved):
LGTM. Thanks for fixing it so fast!

xsoufiane (approved):
+1

nikola-mladenovic (commented):
👍

PR #900, opened 2022-08-08
Migrate sangria-derivation to Scala 3

jchyb:
This includes:
* moving Scala 2 macro to a separate directory
* removing existential types for Scala 3 version of sangria/macros/derive/annotation.scala
* rewriting macro for Scala 3
* replacing now-keyword names, like "enum" in tests
* moving some test classes outside of the scalatest scope to avoid bugs (in StreamSpec.scala)
* changing tests to not expect specific order for the generated fields, as I don't think matching that in Scala 3 to the Scala 2 version is possible 

Last time I mentioned some issues but I managed to find some workarounds for the what I presume was a bug, so now all of the tests pass.


yanns (approved):
This is looking very good! 🏆 

> changing tests to not expect specific order for the generated fields, as I don't think matching that in Scala 3

Does Scala 3 ensure some ordering? And it's not the same as in Scala 2?

modules/derivation/src/main/scala-3/sangria/macros/derive/DeriveObjectTypeMacro.scala
```diff
@@ -0,0 +1,866 @@
...
+            tpe
+
+        NormalArg(name, targetType, expr, optional)
+    }
+
+  private def findKnownMembers[T](tpe: Type[T], includeMethods: Set[String]): List[KnownMember] =
+    import globalQuotes.reflect._
+    given Type[T] = tpe
+    val sym = TypeRepr.of[T].typeSymbol
+    (sym.fieldMembers ++ sym.methodMembers).flatMap { m =>
```
jchyb, line 464: The problem lies here, in Scala 2 version of the macro field members and method members are both obtained through one method, here as far as I know that is not possible so we end up messing with the order just by concatenating symbols like this. Moreover, even in tests which used only fields (or only methods), I could not always match the ordering, even reversing this list. I am unsure if if Scala 3 has their own guarantees for ordering as part of those methods, but I would imagine so

yanns: Thanks for the info!
I suggest merging this PR, and looking back at the ordering if it's an issue later.

PR #904, opened 2022-08-18
Allow ListType to work with any Iterable

stefanolande:


yanns:
Thanks @stefanolande for your contribution. Do you need any help?

stefanolande:
Thanks @yanns . Fixed an issue with the `Resolver` and added a unit test. It should be ok now.

yanns:
This change breaks some binary compatibility. You can add an exception like shown in https://github.com/sangria-graphql/sangria/runs/7916014423?check_suite_focus=true#step:5:58

yanns:
@stefanolande please avoid force-push once the PR is opened. Instead, add commits.

stefanolande:
Added the filter. I'll avoid force pushes from now on 😄

yanns:
You can use `mimaReportBinaryIssues` locally to find out all exceptions that need to be added.

yanns (approved):
💯

Issue #434, opened 2019-01-16
argOpt returns unexpected type at runtime


cneijenhuis:
Given an argument like this:

```scala
val OptionalStringArg: Argument[Option[String]] = Argument("foo", OptionInputType(StringType))
```

I'd expect that argOpt returns `Option[Option[String]]`. The compiler also thinks so, however at runtime I get `Option[String]`.

```scala
val t1: Option[Option[String]] = ctx.argOpt(OptionalStringArg)
val t2: Option[String] = ctx.argOpt(OptionalStringArg) // Does not compile

println(t1) // prints "Some(foobar)", instead of the expected "Some(Some(foobar))"
```

I believe this `instanceOf` is to blame: https://github.com/sangria-graphql/sangria/blob/master/src/main/scala/sangria/schema/Context.scala#L301

paulpdaniels:
I'm not sure that type of argument makes much sense in GraphQL, what would the schema for an option of an option of a string look like?

cneijenhuis:
Right, it looks a little strange without context 😉 I wanted to use this inside a middleware. The middleware is called for any field, including fields that don't have this argument declared in the schema. Therefore:

* `None` - the field does not have the argument
* `Some(None)` - the field has the argument, but it was not set

I don't really need that distinction, but I can't call `arg()` instead of `argOpt()` because it will throw an exception in the first case.

My workaround is to first check if the field has the argument at all, and only if so, call `arg()`.

PR #914, opened 2022-09-27
reproduce issue 434 + optimize code

yanns:


yanns:
can reproduce https://github.com/sangria-graphql/sangria/issues/434, but the fix is not easy.

Issue #671, opened 2021-06-02
[Feature Request] Ability to add directives


Yomanz:
I can't see any way to create any directives within a schema.

I tried but I ran into the same issue that has been open for a while.

https://github.com/sangria-graphql/sangria/issues/467
Any help is greatly appreciated.

xsoufiane:
I might help you here 🤔 there are two parts to consider while using a custom directive. The first part is the definition, for that you use sangria.schema.Directive, as can be seen [here](https://github.com/sangria-graphql/sangria-federated/blob/main/core/src/main/scala/sangria/federation/Directives.scala). The second part is to actually use the defined directive, and for that you need to use sangria.ast.Directive, [here](https://github.com/sangria-graphql/sangria-federated/blob/55787dc575bc36ddaa9dc251cc1d9e829f374472/example/review/src/main/scala/review/State.scala#L16).

Yomanz:
> 
> 
> I might help you here 🤔 there are two parts to consider while using a custom directive. The first part is the definition, for that you use sangria.schema.Directive, as can be seen [here](https://github.com/sangria-graphql/sangria-federated/blob/main/core/src/main/scala/sangria/federation/Directives.scala). The second part is to actually use the defined directive, and for that you need to use sangria.ast.Directive, [here](https://github.com/sangria-graphql/sangria-federated/blob/55787dc575bc36ddaa9dc251cc1d9e829f374472/example/review/src/main/scala/review/State.scala#L16).

I tried this and it didn't seem to apply them to the schema sadly.

yanns:
https://github.com/sangria-graphql/sangria/discussions/913

yanns:
Fixed by https://github.com/sangria-graphql/sangria/pull/918

PR #918, opened 2022-10-10
simple API to add custom directives

yanns:
Context: https://github.com/sangria-graphql/sangria/discussions/913


yanns:
fix https://github.com/sangria-graphql/sangria/issues/671

PR #956, opened 2023-02-06
Fix interface introspection response

nikola-mladenovic:
Steps to reproduce:

```
{
  __schema {
    types {
      name
      interfaces {
        name
      }
    }
  }
}
```

returns:
```
...
{ "name" : "SomeInterfaceName", "interfaces" : null } 
...
```

but according to specs it should return:
```
...
{ "name" : "SomeInterfaceName", "interfaces" : [] } 
...
```

More details: https://spec.graphql.org/October2021/#sec-The-__Type-Type.Interface
> interfaces must return the set of interfaces that an object implements (if none, interfaces must return the empty set).

Knowing that sangria does not support interfaces of interfaces, I have tried to only change this for introspection, so it renders an empty array instead of null.

However all the changes in this PR did not change it (because the following test is still passing:  https://github.com/sangria-graphql/sangria/blob/main/modules/core/src/test/scala/sangria/execution/UnionInterfaceSpec.scala#L119)

@yanns do you have some hints as the most active contributor? 🤔  Any help is appreciated!

### DIFF
--- a/modules/ast/src/main/scala/sangria/ast/QueryAst.scala
+++ b/modules/ast/src/main/scala/sangria/ast/QueryAst.scala
@@ -534,7 +534,8 @@ case class InterfaceTypeDefinition(
     description: Option[StringValue] = None,
     comments: Vector[Comment] = Vector.empty,
     trailingComments: Vector[Comment] = Vector.empty,
-    location: Option[AstLocation] = None)
+    location: Option[AstLocation] = None,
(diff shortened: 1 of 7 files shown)


yanns:
I'd suggest to add a test, as simple as possible, doing an introspection on an object type to check https://spec.graphql.org/October2021/#sec-The-__Type-Type.Object
I'd ignore the interfaces of interface for the moment.

PR #960, opened 2023-02-09
'interfaces' on interface should be an empty list

yanns:
Follow-up of https://github.com/sangria-graphql/sangria/pull/956

Steps to reproduce:

```
{
  __schema {
    types {
      name
      interfaces {
        name
      }
    }
  }
}
```

returns:
```
...
{ "name" : "SomeInterfaceName", "interfaces" : null } 
...
```

but according to specs it should return:
```
...
{ "name" : "SomeInterfaceName", "interfaces" : [] } 
...
```

More details: https://spec.graphql.org/October2021/#sec-The-__Type-Type.Interface
> interfaces must return the set of interfaces that an object implements (if none, interfaces must return the empty set).

Knowing that [sangria does not support interfaces of interfaces for now](https://github.com/sangria-graphql/sangria/pull/597), we just change the introspection result to render an empty array instead of null.

### DIFF
--- a/modules/core/src/main/scala/sangria/introspection/package.scala
+++ b/modules/core/src/main/scala/sangria/introspection/package.scala
@@ -272,7 +272,7 @@ package object introspection {
           "interfaces",
           OptionType(ListType(__Type)),
           resolve = _.value._2 match {
-            case t: ObjectType[_, _] =>
+            case t: ObjectLikeType[_, _] =>
               Some(t.allInterfaces.asInstanceOf[Vector[Type]].map(true -> _))
             case _ => None
           }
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/sangria/introspection/package.scala
```diff
@@ -272,7 +272,7 @@ package object introspection {
           "interfaces",
           OptionType(ListType(__Type)),
           resolve = _.value._2 match {
-            case t: ObjectType[_, _] =>
+            case t: ObjectLikeType[_, _] =>
```
yanns, line 275: `ObjectLikeType` regroups `ObjectType` and `InterfaceType`

PR #966, opened 2023-02-22
Values of correct type when validating SDL. Fixes #965

### LINKED ISSUE #965: Query/Schema validation for default values on object type field arguments not captured (ValueOfCorrectType rule)
From the documentation https://sangria-graphql.github.io/learn/#query-validation 

> Query validation can also be used for SDL validation. ...

Regarding documentation, I expected the following validation to fail, but it is passing:

```scala  
// test can be added to sangria.validation.rules.ValuesOfCorrectTypeSpec
"Type field arguments default values" should {
    "fail when wrong field name" in {

      val Success(doc) = QueryParser.parse(
        """
          |input A {
          | f1: Int
          | f2: Float
          |}
          |""".stripMargin
      )
      val stubSchema = Schema.buildStubFromAst(doc)
      val Success(docUnderTest) = QueryParser.parse(
        """
          |type Test {
          | f(a: A = {wrong: 50}): String
          |}
          |""".stripMargin
      )
      val violations = validator(defaultRule.toList).validateQuery(stubSchema, docUnderTest)
      violations shouldNot be(empty) // but it is empty
      violations.head shouldBe a[UnknownFieldViolation] 

    }
  }
```

I have spent some time recently in the library code, and I think the issue is in [TypeInfo.scala](https://github.com/sangria-graphql/sangria/blob/f49d86092324c2e6274e19984225b04f44ad0d01/modules/core/src/main/scala/sangria/validation/TypeInfo.scala#L8) as it does not push in the `inputTypeStack` the input type, which leads [`ctx.typeInfo.inputType`](https://github.com/sangria-graphql/sangria/blob/f49d86092324c2e6274e19984225b04f44ad0d01/modules/core/src/main/scala/sangria/validation/rules/ValuesOfCorrectType.scala#L37) to be `None` and ignore the checking altogether. 

The following change in `TypeInfo.scala` fixes this issue, and I did not get any other failing tests. It does check the InputValueDefinition with defaultValue and adds the type into the stack, which later can be found by the rule.

```scala

  def enter(node: ast.AstNode) = {
    ancestorStack.push(node)

    node match {
      //code omitted
      case ast.InputValueDefinition(_, valueType, Some(_), _, _, _, _) =>
        inputTypeStack.push(schema.getInputType(valueType))
    }
  }
  def leave(node: ast.AstNode) = {
    node match {
      //code omitted
      case ast.InputValueDefinition(_, _, Some(_), _, _, _, _) => 
        inputTypeStack.pop()
    }

    ancestorStack.pop()
  }
```
It is very hard to get around this issue (or at least I could not find one) since all the stacks in this class are private; extending this class does not help. Adding public methods for pop and push to the stacks could be helpful if one wants to add specific validation rules if the default implementation does not handle specific AstNodes.  

I am new to the library, but if you think this would be enough, I can PR this fix and add public methods for pop and push.

maloku:


maloku:
@yanns do you think these tests are because of the changes I made? If I revert my changes, some tests do fail. Am I missing something?

yanns:
For me, all tests are passing on `main`. It's also a prerequisite for any PR to be merged.
When I checkout your PR, `sangria.schema.AstSchemaMaterializerSpec` is failing. So it should be related to your changes.
But don't panic: check how the tests are failing. Maybe we should update the tests if they show an issue.

maloku:
Oh - that was a good catch. My change ended up validating comments which are `ast.StringValue`  with the type of argument which contains the default value. I adopted it now, and the failing test `AstSchemaMaterializerSpec` is passing.

When I run sbt test I get 4 test failed on main - this is the reason I asked because I expected some test to fail:
```
[info] Tests: succeeded 1246, failed 4, canceled 0, ignored 1, pending 0
[info] *** 4 TESTS FAILED ***
[error] Failed tests:
[error]         sangria.marshalling.QueryAstMarshallingSupportSpec
[error]         sangria.macros.LiteralMacroSpec
[error]         sangria.renderer.SchemaRenderSpec
```
Let's see now

yanns:
> When I run sbt test I get 4 test failed on main - this is the reason I asked because I expected some test to fail:
> 
> ```
> [info] Tests: succeeded 1246, failed 4, canceled 0, ignored 1, pending 0
> [info] *** 4 TESTS FAILED ***
> [error] Failed tests:
> [error]         sangria.marshalling.QueryAstMarshallingSupportSpec
> [error]         sangria.macros.LiteralMacroSpec
> [error]         sangria.renderer.SchemaRenderSpec
> ```

Can you check that further? It's not expected.

yanns (approved):
I really appreciate the attention you gave to the tests! 💯 
As I'm not completely sure about the impact of this change, I'll do the following:
- merge this PR
- cut a RC release
- try this RC release (you can also try the RC release if you want)
If everything is good, then we'll go with a stable release.
If we notice an issue, we can check it.

yanns:
I've tried https://github.com/sangria-graphql/sangria/releases/tag/v3.5.3-RC1 on internal applications, and it does not break anything ;)

maloku:
Great! I do not have a production application as I am only learning (in detail) GraphQL and just happened to pick Scala and Sangria for this journey. But I am still at the beginning, so I may come back with additional stuff, who knows? Thank you so much for supporting me.

PR #972, opened 2023-03-01
Update IEEE 754 wikipedia page link

azarichnyi:
In our project which is based on Sangria we generate pages with all types that exists in our schema. Including scalar types from Sangria. 
We have tests that verify links in types descriptions. These tests started to fail with timeout for `http://en.wikipedia.org/wiki/IEEE_floating_point`, so I checked the link of IEEE 754 wikipedia page and issued this PR with up to date link


yanns (approved):
💯

Issue #567, opened 2021-01-21
Support interface to implement other interface


milliondreams:
Instead of opening an old issue (#388 & #389) which was closed since GraphQL spec in 2018 didn't support this feature https://github.com/sangria-graphql/sangria/issues/388#issuecomment-407569977 creating a new issue to start the discussion.

GraphQL Spec now allows (since Jan 2020) interface to implement other interfaces as documented here -
http://spec.graphql.org/draft/#sec-Interfaces.Interfaces-Implementing-Interfaces

This was merged in the spec by this PR - graphql/graphql-spec#373
and in reference JS implementation by this - graphql/graphql-js#778

Would like to know what contributors' thoughts are wrt to supporting this in Sangria. Do we review #389 or the code has changed enough to warrant a new implementation?

yanns:
Someone has to take to time to open a new PR (can be based on https://github.com/sangria-graphql/sangria/pull/389) starting from the last version.

nickhudkins:
Hey @milliondreams we're limited in our contributors right now, but generally speaking, we absolutely would like to keep up with the spec. 

I suspect #389 may be old enough that a rebase / merge would be difficult, though we'd welcome a fresh PR. 

Additionally, I agree with @OlegIlyenko's review that the original PR got a bit unwieldy to review. Anything that can be done to make the change more targeted would be welcomed greatly.

titaniumcoder:
Is there someone already working on that one? Else I would take a look at the old PR and continue from there and see whether I can put together a PR.

yanns:
> Is there someone already working on that one? Else I would take a look at the old PR and continue from there and see whether I can put together a PR.

I'm not aware of anyone working on that.
Thanks for proposing!

titaniumcoder:
I have a question regarding how things are handled in Sangria. In my opinion (and with just copying lot of the examples from the GraphQL JS reference implementation), the following should probably cover most cases. But how do you handle things like the last two situations where in my opinion, the specification right now is not (yet) clear:

```graphql
interface Being {
    name(surname: Boolean): String
}
interface Mammal {
    mother: Mammal
    father: Mammal
}
interface Pet implements Being {
    name(surname: Boolean): String
}

interface Canine implements Mammal & Being {
    name(surname: Boolean): String
    mother: Canine
    father: Canine
}

type Dog implements Being & Pet & Mammal & Canine {
    name(surname: Boolean): String
    nickname: String
    mother: Dog
    father: Dog
}

[...]

PR #981, opened 2023-03-29
Add support for interface implementing interface

filosganga:
This is an initial support for #567 and supersede #597 that is based on very old base

The rendering has been a bit challenging and I took a non-standard path there

### DIFF
--- a/modules/core/src/main/scala/sangria/introspection/package.scala
+++ b/modules/core/src/main/scala/sangria/introspection/package.scala
@@ -274,8 +274,8 @@ package object introspection {
           resolve = _.value._2 match {
             case t: ObjectType[_, _] =>
               Some(t.allInterfaces.asInstanceOf[Vector[Type]].map(true -> _))
-            case _: InterfaceType[_, _] =>
-              Some(Vector.empty)
+            case t: InterfaceType[_, _] =>
+              Some(t.interfaces.asInstanceOf[List[Type]].map(true -> _))
             case _ => None
           }
         ),
--- a/modules/core/src/main/scala/sangria/introspection/package.scala
+++ b/modules/core/src/main/scala/sangria/introspection/package.scala
@@ -272,10 +272,8 @@ package object introspection {
           "interfaces",
           OptionType(ListType(__Type)),
           resolve = _.value._2 match {
-            case t: ObjectType[_, _] =>
+            case t: ObjectLikeType[_, _] =>
               Some(t.allInterfaces.asInstanceOf[Vector[Type]].map(true -> _))
-            case t: InterfaceType[_, _] =>
-              Some(t.interfaces.asInstanceOf[List[Type]].map(true -> _))
             case _ => None
           }
         ),
--- a/modules/core/src/main/scala/sangria/schema/AstSchemaMaterializer.scala
+++ b/modules/core/src/main/scala/sangria/schema/AstSchemaMaterializer.scala
@@ -697,8 +697,9 @@ class AstSchemaMaterializer[Ctx] private (
       tpe,
       extensions.toList,
       () => buildFields(origin, tpe, tpe.fields, extensions).toList,
-      extendInterfaces(origin, tpe, extensions),
-      this)
+      buildInterfaces(origin, tpe, tpe.interfaces, extensions).toList,
+      this
+    )
   }
 
   def extendEnumType(origin: MatOrigin, tpe: EnumType[_]) = {
@@ -740,6 +741,7 @@ class AstSchemaMaterializer[Ctx] private (
       tpe,
       extensions.toList,
       () => extendFields(origin, tpe, extensions),
+      extendInterfaces(origin, tpe, extensions),
       this)
   }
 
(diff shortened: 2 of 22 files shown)


modules/core/src/main/scala/sangria/introspection/package.scala
```diff
@@ -274,8 +274,8 @@ package object introspection {
           resolve = _.value._2 match {
             case t: ObjectType[_, _] =>
               Some(t.allInterfaces.asInstanceOf[Vector[Type]].map(true -> _))
-            case _: InterfaceType[_, _] =>
-              Some(Vector.empty)
+            case t: InterfaceType[_, _] =>
+              Some(t.interfaces.asInstanceOf[List[Type]].map(true -> _))
```
yanns, line 278: Both cases can be covered by `ObjectLikeType`.

yanns (commented):
Very cool to see progress here!

modules/core/src/main/scala/sangria/introspection/package.scala
filosganga: Yep I have just re-read the spec, this is wrong, the transitively implemented interfaces should be ALL added

filosganga: Fixed

yanns: 💯

filosganga:
@yanns This is pretty much done but there are a couple of missing validations that I don't know how to handle:

**All the implementing fields must be declare**
When a type or an interface implement another interface, all the implementing fields must be declared. This is not currently validated even without interfaces implementing interfaces, so can it be addressed in a follow-up PR?

**The implementation hierarchy MUST not include loops**
I have managed to check for interfaces implementing themselves, however is very complex to spot issues like this:

````graphql
interface Foo implement Baz {
  foo: String
}

interface Bar implement Foo{
  bar: String
  foo: String
}

interface Baz implement Bar & Foo {
  baz: String
  bar: String
  foo: String
}
````

Because the materializer/builder goes in SO when trying to resolve the types. Is there a way to validate the ast before running the materializer/builder?

Cheers

**Comment by yanns** [comment]
> **All the implementing fields must be declare** When a type or an interface implement another interface, all the implementing fields must be declared. This is not currently validated even without interfaces implementing interfaces, so can it be addressed in a follow-up PR?

Yes

> **The implementation hierarchy MUST not include loops** I have managed to check for interfaces implementing themselves, however is very complex to spot issues like this:
> 
> ```graphql
> interface Foo implement Baz {
>   foo: String
> }
> 
> interface Bar implement Foo{
>   bar: String
>   foo: String
> }
> 
> interface Baz implement Bar & Foo {
>   baz: String
>   bar: String
>   foo: String
> }
> ```
> 
> Because the materializer/builder goes in SO when trying to resolve the types. Is there a way to validate the ast before running the materializer/builder?

I'm not sure. I'm thinking of https://github.com/sangria-graphql/sangria/pull/966/files were some validation is done on the AST directly. Does this help?

**Comment by filosganga** [comment]
> > **All the implementing fields must be declare** When a type or an interface implement another interface, all the implementing fields must be declared. This is not currently validated even without interfaces implementing interfaces, so can it be addressed in a follow-up PR?
> 
> Yes
> 
> > **The implementation hierarchy MUST not include loops** I have managed to check for interfaces implementing themselves, however is very complex to spot issues like this:
> > ```graphql
> > interface Foo implement Baz {
> >   foo: String
> > }
> > 
> > interface Bar implement Foo{
> >   bar: String
> >   foo: String
> > }
> > 
> > interface Baz implement Bar & Foo {
> >   baz: String
> >   bar: String
> >   foo: String
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Because the materializer/builder goes in SO when trying to resolve the types. Is there a way to validate the ast before running the materializer/builder?
> 
> I'm not sure. I'm thinking of https://github.com/sangria-graphql/sangria/pull/966/files were some validation is done on the AST directly. Does this help?

Yep, I think so.

Also, thank you to have allowed the workflow to run, I need to fix a few things :)

**Comment by yanns** [comment]
```
[error] (sangria-core / Compile / scalafmtCheck) scalafmt: 1 files must be formatted (/home/runner/work/sangria/sangria/modules/core)
```

**Comment by filosganga** [comment]
@yanns This is ready to review. However, it is impossible to retain binary compatibility because Sangria uses case classes with default values. I have added the mima filters but I think a major version bump may be needed

> > All the implementing fields must be declare When a type or an interface implement another interface, all the implementing fields must be declared. This is not currently validated even without interfaces implementing interfaces, so can it be addressed in a follow-up PR?
> 
> Yes
Actually, I saw that in (https://github.com/sangria-graphql/sangria/blob/7970fa7be466d7bbd5ac07e699e68d769144bc17/modules/core/src/main/scala/sangria/schema/SchemaValidationRule.scala#L125-L127) Sangria allows to avoid defining all the inherited fields.

I have implemented cycle detection in the AST before being materialised

**Comment by yanns** [comment]
Really cool, I'll have a look at this.
Yes, I think I'll make a major version bump.

File: modules/core/src/main/scala/sangria/schema/AstSchemaMaterializer.scala [hunk 2]
```diff
@@ -296,12 +301,45 @@ class AstSchemaMaterializer[Ctx] private (
...
+      interfaceType: ast.InterfaceTypeDefinition): Vector[Violation] = {
+
+    def go(
+        validating: ast.InterfaceTypeDefinition,
+        visited: Set[ast.InterfaceTypeDefinition]): Vector[Violation] = {
+
+      val allInterfaceDefs: Vector[ast.InterfaceTypeDefinition] = (interfaceTypeExtensionDefs
+        .find(ie => ie.name == validating.name)
+        .map(ie => ie.interfaces)
+        .getOrElse(Vector.empty) ++ validating.interfaces).flatMap { i =>
```
yanns, line 321: nitpick: using `.fold(Vector.empty[NamedType])(ie => ie.interfaces) ` instead of map and getOrElse

yanns (approved):
Beautiful PR, with lot of tests! 💯

yanns:
> If you do a major version bump, I can remove the apply methods that meant to keep source compatibility or limit the changes.

Good point.
We should still try not to break user code if possible. It's sometimes feasible with default values. But sometimes we need those new apply methods.

yanns:
For reference, this is in the spec since https://github.com/graphql/graphql-spec/blob/main/changelogs/October2021.md

modules/core/src/main/scala/sangria/schema/AstSchemaMaterializer.scala
filosganga: Fixed

filosganga:
> Good point. We should still try not to break user code if possible. It's sometimes feasible with default values. But sometimes we need those new apply methods.

Fair point

yanns:
released in https://github.com/sangria-graphql/sangria/releases/tag/v4.0.0-RC1

PR #989, opened 2023-03-31
Fix TypeComparators

filosganga:
Fix the isSubType of typeComparator and add specific tests

### DIFF
--- a/modules/core/src/main/scala/sangria/schema/Schema.scala
+++ b/modules/core/src/main/scala/sangria/schema/Schema.scala
@@ -1529,10 +1529,33 @@ case class Schema[Ctx, Val](
 
     directImplementations.map { case (name, directImpls) =>
       name -> directImpls.flatMap(findDescendants).groupBy(_.name).map(_._2.head).toVector
-    }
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/sangria/schema/Schema.scala
```diff
@@ -1529,10 +1529,33 @@ case class Schema[Ctx, Val](
 
     directImplementations.map { case (name, directImpls) =>
       name -> directImpls.flatMap(findDescendants).groupBy(_.name).map(_._2.head).toVector
-    }
+    } ++ unionTypes.values.map(ut => ut.name -> ut.types.toVector)
```
filosganga, line 1532: Actually, I don't like this

yanns (approved):

modules/core/src/main/scala/sangria/schema/Schema.scala
yanns: Do you have a better idea?

filosganga: Yeah I have changed back and use union only in possibleTypes

Issue #982, opened 2023-03-30
support @deprecated on argument and input field


yanns:
In the draft spec, the `@deprecated` directive can now be also used on arguments and input fields.

[Draft](https://spec.graphql.org/draft/#sec--deprecated):
```
directive @deprecated(
  reason: String = "No longer supported"
) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
```

vs [last spec from October 2021](https://spec.graphql.org/October2021/#sec--deprecated):
```
directive @deprecated(
  reason: String = "No longer supported"
) on FIELD_DEFINITION | ENUM_VALUE
```

[RFC](https://github.com/graphql/graphql-spec/pull/525)

PR #992, opened 2023-04-09
Support @deprecated on argument and input field

yanns:
Fixes https://github.com/sangria-graphql/sangria/issues/982


yanns:
https://github.com/sangria-graphql/sangria/pull/1063

PR #993, opened 2023-04-14
Call scalar coerceUserInput before coerceOutput

filosganga:


modules/core/src/main/scala/sangria/execution/FutureResolver.scala
```diff
@@ -1405,7 +1405,11 @@ private[execution] class FutureResolver[Ctx](
             if (isUndefinedValue(value))
               None
             else {
-              val coerced = scalar.coerceOutput(value, marshaller.capabilities)
+              val corcedInput = scalar
+                .coerceUserInput(value)
+                // Do we need a specific exepction here?
+                .fold(e => throw new ValidationError(Vector(e), exceptionHandler), identity)
+              val coerced = scalar.coerceOutput(corcedInput, marshaller.capabilities)
```
filosganga, line 1412: Throwing an exception here because the whole block is protected by try/catch

modules/core/src/test/scala/sangria/schema/ResolverBasedAstSchemaBuilderSpec.scala
```diff
@@ -34,6 +34,7 @@ class ResolverBasedAstSchemaBuilderSpec extends AnyWordSpec with Matchers with F
       "UUID",
       coerceOutput = (v, _) => v.toString,
       coerceUserInput = {
+        case uuid: UUID => Right(uuid)
```
yanns, line 37: I try to understand:
`coerceUserInput` is now called twice:
- the first time to parse the user input as `String` into a `UUID`
- the second time to confirm that a `UUID` is a valid value
?

filosganga: No it is not called, twice, but I have spent a bit of time looking at Apollo implementation, and I think, I would just need to change the scalars rather than Sangria.

In Apollo the scalar are represented by 3 functions:
- `serialize` it is the same as `coerceOutput`, it coerces the scalar to a JSON representation
- `parseLiteral` it is the same as `coerceInput` that is bit misleading, it basically coerces and hardcoded value into the scalar type
- `parseValue` it is the same as `coerceUserInput` that basically coerces the values coming from the users via not hardcoded variables.

When you have an untyped underlying model, like JSON, the `serialize` function should take care **also** of parsing the type from the underlying model.

PR #997, opened 2023-04-28
Add scaladoc for Arg methods

keirlawson:
I got tripped up by the semantics of these, so thought I would contribute back some docs


yanns (approved):

yanns:
Thanks!

PR #1001, opened 2023-05-16
Allow overriding builtin scalars

filosganga:
This PR reverses the logic of getNamedType to allow overriding builtin scalars


yanns (approved):
This is looking good. Is the PR ready?

filosganga:
> This is looking good. Is the PR ready?

Thank you @yanns 

Yes it is ready

PR #1017, opened 2023-06-12
Add an option to limit number of errors returned

bc-dima-pasieka:
# What?
Add an option to limit the number of errors returned.

# Why?
There is a possibility that the lack of limits could be abused to induce memory pressure in the server/service handling the request and can lead to the potential denial of service (i.e. no DDoS needed).

### DIFF
--- a/modules/core/src/main/scala/sangria/validation/QueryValidator.scala
+++ b/modules/core/src/main/scala/sangria/validation/QueryValidator.scala
@@ -45,14 +45,18 @@ object QueryValidator {
     new SingleFieldSubscriptions
   )
 
-  def ruleBased(rules: List[ValidationRule]) = RuleBasedQueryValidator(rules)
+  @deprecated("use ruleBased setting 'errorsLimit' instead", "4.0.1")
+  def ruleBased(rules: List[ValidationRule]): RuleBasedQueryValidator =
+    RuleBasedQueryValidator(rules)
+  def ruleBased(rules: List[ValidationRule], errorsLimit: Option[Int]): RuleBasedQueryValidator =
+    new RuleBasedQueryValidator(rules, errorsLimit)
 
-  val empty = new QueryValidator {
+  val empty: QueryValidator = new QueryValidator {
     def validateQuery(schema: Schema[_, _], queryAst: ast.Document): Vector[Violation] =
       Vector.empty
   }
 
-  val default: RuleBasedQueryValidator = ruleBased(allRules)
+  val default: RuleBasedQueryValidator = ruleBased(allRules, errorsLimit = Some(10))
 }
 
 class RuleBasedQueryValidator(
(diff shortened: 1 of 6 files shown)


modules/core/src/main/scala/sangria/validation/QueryValidator.scala
```diff
@@ -148,7 +156,8 @@ class ValidationContext(
     val schema: Schema[_, _],
     val doc: ast.Document,
     val sourceMapper: Option[SourceMapper],
-    val typeInfo: TypeInfo) {
+    val typeInfo: TypeInfo,
+    errorsLimit: Option[Int] = None) {
```
yanns, line 160: Making errorsLimit immutable as `ValidationContext` is a call.
```suggestion
    val errorsLimit: Option[Int] = None) {
```

bc-dima-pasieka: omitting `val` will make it private to the class since it's not a case class, won't it? 🤔 Do we want to make it public?

bc-dima-pasieka:
> apply(rules: List[ValidationRule]): RuleBasedQueryValidator in the companion object

It looks like it will be a breaking change if we do it this way - we will not be able to call `new RuleBasedQueryValidator(newRules)` anymore (only `RuleBasedQueryValidator(newRules)`)
```
object RuleBasedQueryValidator {
  def apply(rules: List[ValidationRule]): RuleBasedQueryValidator = RuleBasedQueryValidator(rules, None)
}
```
The same with `ValidationContext`. 

🤔 Would we be fine introducing breaking change?

modules/core/src/main/scala/sangria/validation/QueryValidator.scala
yanns: oh yes sorry. Keep it like this.

yanns:
`RuleBasedQueryValidator` is more likely being called from user.
Sure that we cannot keep the `new RuleBasedQueryValidator(rules)`, but let's try to keep the `apply` method.

For `ValidationContext`, I think it's ok to make a breaking change. It's unlikely that users initiate it themselves.
I'll add this to the release notes.

bc-dima-pasieka:
@yanns let me know if it looks better now 🙌

yanns:
It's looking good, thanks!
I'll take care of adding exceptions for the rest of binary incompatibilities,

yanns:
I propose to set a default limit to make default usage safe: https://github.com/sangria-graphql/sangria/pull/1017/commits/aab5f5bc530d7cec547ad1592eba5f0db45ec913
WDYT? 10 errors are sufficient?

yanns:
I propose to set a default limit to make default usage safe: https://github.com/sangria-graphql/sangria/pull/1017/commits/aab5f5bc530d7cec547ad1592eba5f0db45ec913
WDYT? 10 errors are sufficient?

bc-dima-pasieka:
10 sounds like a good default value! 👍

PR #1022, opened 2023-06-30
DRAFT / WIP: Limit number of errors returned (when variables are used)

bc-dima-pasieka:
Follow up to https://github.com/sangria-graphql/sangria/pull/1017

The solution from the previous PR works only for cases where input is specified (hardcoded) as a part of the query but doesn't work/executed when variables are used. For example, the following query will still return an unlimited number of errors:
```
mutation createProduct($input: CreateProductInput!) {
  product {
    createProduct(input: $input) {
      ...
    }
  }
}

// variables
{
    "input":{
      "field_with_invalid_objects":[{},{},{},{},{}, ...]
    }
  }
```

@yanns I'd appreciate your guidance here on the best way to design it 🙂

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/Executor.scala
+++ b/modules/core/src/main/scala/sangria/execution/Executor.scala
@@ -4,9 +4,10 @@ import sangria.ast
 import sangria.ast.SourceMapper
 import sangria.marshalling.{InputUnmarshaller, ResultMarshaller}
 import sangria.schema._
-import sangria.validation.QueryValidator
+import sangria.validation.{QueryValidator, RuleBasedQueryValidator}
 import InputUnmarshaller.emptyMapVars
 import sangria.execution.deferred.DeferredResolver
+
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
@@ -43,7 +44,9 @@ case class Executor[Ctx, Root](
         userContext,
         exceptionHandler,
         scalarMiddleware,
-        false)(um)
+        false,
+        errorsLimit = queryValidator.errorsLimit
+      )(um)
 
       val executionResult = for {
         operation <- Executor.getOperation(exceptionHandler, queryAst, operationName)
@@ -155,7 +158,9 @@ case class Executor[Ctx, Root](
         userContext,
         exceptionHandler,
         scalarMiddleware,
-        false)(um)
+        false,
+        errorsLimit = queryValidator.errorsLimit
+      )(um)
 
       val executionResult = for {
         operation <- Executor.getOperation(exceptionHandler, queryAst, operationName)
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/sangria/execution/Executor.scala
```diff
@@ -43,7 +44,9 @@ case class Executor[Ctx, Root](
         userContext,
         exceptionHandler,
         scalarMiddleware,
-        false)(um)
+        false,
+        errorsLimit = queryValidator.errorsLimit
```
bc-dima-pasieka, line 48: The idea was to define errorsLimit only once (e.g. on the queryValidator) and re-use in other places

modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
```diff
@@ -580,98 +582,108 @@ class ValueCoercionHelper[Ctx](
...
-    case (OptionInputType(_), _) => Vector.empty
-    case (_, None) => Vector(NotNullValueIsNullViolation(sourceMapper, Nil))
-
-    case (ListInputType(ofType), Some(values)) if um.isListNode(values) =>
-      um.getListValue(values)
-        .toVector
-        .flatMap(v =>
+  def isValidValue[In](tpe: InputType[_], input: Option[In], errors: ListBuffer[Violation])(implicit
+      um: InputUnmarshaller[In]): Vector[Violation] = {
+    // TODO: NEED TO ADD VALUES TO `errors` to make it work properly
```
bc-dima-pasieka, line 587: This code doesn't work correctly because I don't mutate errors buffer, but this is one of the options how it can be implemented

modules/core/src/main/scala/sangria/execution/ValueCollector.scala
```diff
@@ -63,7 +64,14 @@ class ValueCollector[Ctx, Input](
...
 
-      val (errors, values) = res.partition(_._2.isLeft)
+      val (allErrors, values) = res.partition(_._2.isLeft)
+
+      // additionally reduce the number of errors
+      // although we limit the number of errors per input, it's still possible to exceed the limit if errors are combined.
+      val errors = errorsLimit match {
+        case Some(limit) => allErrors.take(limit)
+        case _ => allErrors
+      }
```
bc-dima-pasieka, line 74: it seems like there are other places where violations could be returned, but I'm not sure if they should be updated as well.

modules/core/src/main/scala/sangria/validation/QueryValidator.scala
```diff
@@ -12,6 +12,7 @@ import scala.reflect.{ClassTag, classTag}
 
 trait QueryValidator {
   def validateQuery(schema: Schema[_, _], queryAst: ast.Document): Vector[Violation]
+  def errorsLimit: Option[Int]
```
bc-dima-pasieka, line 15: Making it public so it can be re-used?

yanns:
Thanks for PR.
I'm not very familiar with this code. What kind of help do you need?

bc-dima-pasieka:
> I'm not very familiar with this code
Got it! I was looking for some guidance on the best way/place to fix it. I guess I will need to dig deeper into the logic then 😃

bc-dima-pasieka:
Closing in favor of https://github.com/sangria-graphql/sangria/pull/1034

PR #1034, opened 2023-07-17
Limit number of errors returned (when variables are used)

bc-dima-pasieka:
Follow up to https://github.com/sangria-graphql/sangria/pull/1017

The solution from the previous PR works only for cases where input is specified (hardcoded) as a part of the query but doesn't work/executed when variables are used. For example, the following query will still return an unlimited number of errors:
```
mutation createProduct($input: CreateProductInput!) {
  product {
    createProduct(input: $input) {
      ...
    }
  }
}

// variables
{
    "input":{
      "field_with_invalid_objects":[{},{},{},{},{}, ...]
    }
  }
```

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/Executor.scala
+++ b/modules/core/src/main/scala/sangria/execution/Executor.scala
@@ -19,7 +19,8 @@ case class Executor[Ctx, Root](
     deprecationTracker: DeprecationTracker = DeprecationTracker.empty,
     middleware: List[Middleware[Ctx]] = Nil,
     maxQueryDepth: Option[Int] = None,
-    queryReducers: List[QueryReducer[Ctx, _]] = Nil
+    queryReducers: List[QueryReducer[Ctx, _]] = Nil,
+    errorsLimit: Option[Int] = None
 )(implicit executionContext: ExecutionContext) {
   def prepare[Input](
       queryAst: ast.Document,
@@ -29,7 +30,7 @@ case class Executor[Ctx, Root](
       variables: Input = emptyMapVars
   )(implicit um: InputUnmarshaller[Input]): Future[PreparedQuery[Ctx, Root, Input]] = {
     val (violations, validationTiming) =
-      TimeMeasurement.measure(queryValidator.validateQuery(schema, queryAst))
+      TimeMeasurement.measure(queryValidator.validateQuery(schema, queryAst, errorsLimit))
 
     if (violations.nonEmpty)
       Future.failed(ValidationError(violations, exceptionHandler))
@@ -49,7 +50,9 @@ case class Executor[Ctx, Root](
         operation <- Executor.getOperation(exceptionHandler, queryAst, operationName)
         unmarshalledVariables <- valueCollector.getVariableValues(
           operation.variables,
-          scalarMiddleware)
+          scalarMiddleware,
+          errorsLimit
+        )
         fieldCollector = new FieldCollector[Ctx, Root](
           schema,
           queryAst,
@@ -141,7 +144,7 @@ case class Executor[Ctx, Root](
       um: InputUnmarshaller[Input],
       scheme: ExecutionScheme): scheme.Result[Ctx, marshaller.Node] = {
     val (violations, validationTiming) =
-      TimeMeasurement.measure(queryValidator.validateQuery(schema, queryAst))
+      TimeMeasurement.measure(queryValidator.validateQuery(schema, queryAst, errorsLimit))
 
     if (violations.nonEmpty)
       scheme.failed(ValidationError(violations, exceptionHandler))
@@ -161,7 +164,9 @@ case class Executor[Ctx, Root](
         operation <- Executor.getOperation(exceptionHandler, queryAst, operationName)
         unmarshalledVariables <- valueCollector.getVariableValues(
           operation.variables,
-          scalarMiddleware)
+          scalarMiddleware,
+          errorsLimit
+        )
         fieldCollector = new FieldCollector[Ctx, Root](
           schema,
           queryAst,
@@ -324,7 +329,8 @@ object Executor {
       deprecationTracker: DeprecationTracker = DeprecationTracker.empty,
       middleware: List[Middleware[Ctx]] = Nil,
       maxQueryDepth: Option[Int] = None,
-      queryReducers: List[QueryReducer[Ctx, _]] = Nil
+      queryReducers: List[QueryReducer[Ctx, _]] = Nil,
+      errorsLimit: Option[Int] = None
   )(implicit
       executionContext: ExecutionContext,
       marshaller: ResultMarshaller,
@@ -338,7 +344,8 @@ object Executor {
       deprecationTracker,
       middleware,
       maxQueryDepth,
-      queryReducers)
+      queryReducers,
+      errorsLimit)
       .execute(queryAst, userContext, root, operationName, variables)
 
   def prepare[Ctx, Root, Input](
@@ -354,7 +361,8 @@ object Executor {
       deprecationTracker: DeprecationTracker = DeprecationTracker.empty,
       middleware: List[Middleware[Ctx]] = Nil,
       maxQueryDepth: Option[Int] = None,
-      queryReducers: List[QueryReducer[Ctx, _]] = Nil
+      queryReducers: List[QueryReducer[Ctx, _]] = Nil,
+      errorsLimit: Option[Int] = None
   )(implicit
       executionContext: ExecutionContext,
       um: InputUnmarshaller[Input]): Future[PreparedQuery[Ctx, Root, Input]] =
@@ -366,7 +374,8 @@ object Executor {
       deprecationTracker,
       middleware,
       maxQueryDepth,
-      queryReducers)
+      queryReducers,
+      errorsLimit)
       .prepare(queryAst, userContext, root, operationName, variables)
 
   def getOperationRootType[Ctx, Root](
--- a/modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
+++ b/modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
@@ -580,8 +580,10 @@ class ValueCoercionHelper[Ctx](
     nodeLocation.toList ++ firstValue.toList
   }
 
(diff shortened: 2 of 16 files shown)


modules/core/src/main/scala/sangria/execution/Executor.scala
```diff
@@ -19,7 +19,8 @@ case class Executor[Ctx, Root](
     deprecationTracker: DeprecationTracker = DeprecationTracker.empty,
     middleware: List[Middleware[Ctx]] = Nil,
     maxQueryDepth: Option[Int] = None,
-    queryReducers: List[QueryReducer[Ctx, _]] = Nil
+    queryReducers: List[QueryReducer[Ctx, _]] = Nil,
+    errorsLimit: Option[Int] = None
```
bc-dima-pasieka, line 23: The idea is to pass `errorsLimit` to `execute/prepare` methods and it will propagate it further to the query validator and to the variables validator.

modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
```diff
@@ -580,107 +580,126 @@ class ValueCoercionHelper[Ctx](
...
-            SchemaRenderer.renderTypeName(objTpe, true),
-            f,
-            sourceMapper,
-            Nil)
-      }
+  def isValidValue[In](inputType: InputType[_], input: Option[In], errorsLimit: Option[Int])(
+      implicit um: InputUnmarshaller[In]): Vector[Violation] = {
+
+    // keeping track of the number of errors
+    var errors = 0
```
bc-dima-pasieka, line 587: Using mutable variable was the only viable option here to avoid making huge refactoring and keeping the same signature of the original method (by returning Vector[Violation]).

The alternative 1 was to use ListBuffer[Violation] but it would require making signature changes (probably just to return Unit) + it didn't work well with `.map(ListValueViolation(0, _, sourceMapper, Nil))` if we I did it.

The alternative 2 was to propagate the number of errors back to the recursion function but I didn't make it work properly 🙂

modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
```diff
@@ -580,107 +580,126 @@ class ValueCoercionHelper[Ctx](
...
+    def addViolation(violation: Violation): Vector[Violation] = {
+      errors += 1
+      Vector(violation)
+    }
 
-      val fieldViolations =
-        objTpe.fields.toVector.flatMap(f =>
-          isValidValue(f.fieldType, um.getMapValue(valueMap, f.name))
-            .map(MapValueViolation(f.name, _, sourceMapper, Nil)))
+    def isValidValueRec(tpe: InputType[_], in: Option[In])(implicit
```
bc-dima-pasieka, line 593: I created a separate recursion function (basically the old method migrated here for the most part) to be called internally so it has easy access to mutable variable

modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
```diff
@@ -580,107 +580,126 @@ class ValueCoercionHelper[Ctx](
...
-          Nil))
+            fieldViolations ++ unknownFields
 
-    case (scalar: ScalarType[_], Some(value)) if um.isScalarNode(value) =>
-      val coerced = um.getScalarValue(value) match {
-        case node: ast.Value => scalar.coerceInput(node)
-        case other => scalar.coerceUserInput(other)
-      }
+          case (objTpe: InputObjectType[_], _) =>
+            addViolation(
```
bc-dima-pasieka, line 642: So instead of returning Vector(violation), we call `addViolation(violation)` that mutates number of errors and returns the same Vector(violation)

modules/core/src/main/scala/sangria/validation/QueryValidator.scala
```diff
@@ -63,13 +71,19 @@ class RuleBasedQueryValidator(
...
+      schema: Schema[_, _],
+      queryAst: ast.Document,
+      errorsLimitOverride: Option[Int]): Vector[Violation] = {
     val ctx = new ValidationContext(
       schema,
       queryAst,
       queryAst.sourceMapper,
       new TypeInfo(schema),
-      errorsLimit)
+      errorsLimitOverride)
```
bc-dima-pasieka, line 86: It would take precedence over RuleBasedQueryValidator.errorsLimit so we could pass and use a single limit via `execute` method

bc-dima-pasieka:
@yanns pls take a look when you have time 🙂

modules/core/src/main/scala/sangria/execution/Executor.scala
yanns: As an alternative, we could use the field `errorsLimit` on `queryValidator`.
I tried it quickly in https://github.com/bc-dima-pasieka/sangria/pull/1

bc-dima-pasieka: Yeah, that's how I implemented it initially https://github.com/sangria-graphql/sangria/pull/1022, but it seemed a bit weird to use the property of queryValidator to validate something but query 🤔

yanns: I understand. I'll check again in the light of this new knowledge.
I was also trying not to break binary compatibility once again.

yanns:
so that you're not surprised by the delay:
I'll have some time off. I'll start the review again after.

bc-dima-pasieka:
Hey @yanns, just following up in case you are back 🙂

bc-dima-pasieka:
@yanns 

> Do you think that we should revert part of the initial PR, and go into that direction: [bc-dima-pasieka#2](https://github.com/bc-dima-pasieka/sangria/pull/2) ?

Probably yes, it might play better in the long run. Let me update the PR

modules/core/src/main/scala/sangria/validation/QueryValidator.scala
```diff
@@ -11,7 +11,10 @@ import scala.collection.mutable.{ListBuffer, Map => MutableMap, Set => MutableSe
 import scala.reflect.{ClassTag, classTag}
 
 trait QueryValidator {
-  def validateQuery(schema: Schema[_, _], queryAst: ast.Document): Vector[Violation]
+  def validateQuery(
```
bc-dima-pasieka, line 14: @yanns based on the suggestion from https://github.com/bc-dima-pasieka/sangria/pull/2/files#diff-f36cd7c72002ebd8d7f5fcc1f1b1026e4f54c67bfe9d7268425ba73ecd8a272bL14 - are we ok introducing a breaking change?

yanns: This PR introduces quite a lot of important breaking changes.
I'd suggest that we ignore them for now to focus on the feature.
Once we're happy with the result, we can check how to handle the breaking changes.
You can temporary remove the CI step about breaking changes so that we can focus on the tests.

bc-dima-pasieka: I commented out a few checks. Let me know what would be the best next steps from here. It seems to be "ready" from my end

modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
```diff
@@ -580,107 +580,126 @@ class ValueCoercionHelper[Ctx](
...
-    case (objTpe: InputObjectType[_], Some(valueMap)) if um.isMapNode(valueMap) =>
-      val unknownFields = um.getMapKeys(valueMap).toVector.collect {
-        case f if !objTpe.fieldsByName.contains(f) =>
-          UnknownInputObjectFieldViolation(
-            SchemaRenderer.renderTypeName(objTpe, true),
-            f,
-            sourceMapper,
-            Nil)
-      }
+  def isValidValue[In](inputType: InputType[_], input: Option[In], errorsLimit: Option[Int])(
```
yanns, line 583: We can make this private (we're already breaking binary compatibility anyways)

yanns (commented):
This is looking good to me.
You can add the exceptions for the binary compatibility checks:
```
    mimaBinaryIssueFilters ++= Seq(
      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.execution.Executor.apply"),
      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.execution.Executor.copy"),
      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.execution.Executor.this"),
      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.execution.Executor.execute"),
      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.execution.Executor.prepare"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.execution.QueryReducerExecutor.reduceQueryWithoutVariables"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.execution.ValueCoercionHelper.isValidValue"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.execution.ValueCoercionHelper.getVariableValue"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.execution.batch.BatchExecutor.executeBatch"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.schema.ResolverBasedAstSchemaBuilder.validateSchema"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.validation.QueryValidator.validateQuery"),
      ProblemFilters.exclude[ReversedMissingMethodProblem](
        "sangria.validation.QueryValidator.validateQuery"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.validation.RuleBasedQueryValidator.validateQuery"),
      ProblemFilters.exclude[DirectMissingMethodProblem](
        "sangria.validation.ValidationContext.this")
    ),
```
I've tried to minimize this, but it's a nightmare as the new argument is optional.
The source compatibility is good though.

Can you add those exceptions are enable the binary check again?

yanns (approved):
Thanks a lot!

modules/core/src/main/scala/sangria/execution/Executor.scala
yanns: https://github.com/sangria-graphql/sangria/pull/1049

PR #1063, opened 2023-10-05
add input value deprecation

asteinwedel:
Input value deprecation is a feature added a few years ago elsewhere that we would like to add support for. See [here](https://github.com/graphql/graphql-js/blob/main/src/utilities/getIntrospectionQuery.ts#L34) and [here](https://github.com/graphql/graphql-js/blob/7a6d055defd7a888f50c6133faae2b9010d45fad/src/type/directives.ts#L220) for graphql-js usage in introspection.

There are two main things I believe I still need to work on, one of which particularly I'd like some input on.
- ~~[x] add to deprecation tracking~~ removed
- [x] handling directive arg deprecation in various places~~ removed
- [x] fix a test

~~Regarding directives, when looking at tests in `DocumentAnalyzerSpec` and `IntrospectionSchemaMaterializerSpec`, I would like to test a field having a directive with a deprecated argument. However, the field only has `ast.Directive`s on it, which lack deprecation information (same for other types with deprecation, it's not in the `ast` objects).~~

~~It seems like I'd have to replace `Field.astDirectives: Vector[ast.Directive]` to instead be `Field.directives: Vector[schema.Directive]`. It is more align that way with `Field.arguments: List[schema.Argument]`, which has deprecation information. So it seems more in line with keeping `schema` type objects on itself.~~

~~I was wondering if that sounds like a good approach, or why `Field` has `ast.Directive` instead of `schema.Directive` in the first place?~~

~~(same goes for `InputField`, `Argument`, and various other places directives, being `ast.Directive` instead of `schema.Directive`)~~

answer @ `ast.Directive` vs `schema.Directive`, can usually find the corresponding `schema.Directive` on `Schema.directives`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -116,7 +116,40 @@ lazy val core = project
       ProblemFilters.exclude[DirectMissingMethodProblem](
         "sangria.validation.RuleBasedQueryValidator.validateQuery"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "sangria.validation.ValidationContext.this")
+        "sangria.validation.ValidationContext.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.IntrospectionInputValue.apply"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.IntrospectionInputValue.copy"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.IntrospectionInputValue.this"),
+      ProblemFilters.exclude[MissingTypesProblem]("sangria.introspection.IntrospectionInputValue$"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.IntrospectionInputValue.apply"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.package.introspectionQueryString"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.package.introspectionQuery"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.package.introspectionQuery"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.introspection.package.introspectionQueryString"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.Argument.apply"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.Argument.copy"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+        "sangria.schema.Argument.copy$default$6"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("sangria.schema.Argument._6"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.Argument.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.Argument.apply"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.InputField.apply"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.InputField.copy"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+        "sangria.schema.InputField.copy$default$5"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("sangria.schema.InputField._5"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.InputField.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("sangria.schema.InputField.apply"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "sangria.schema.InputValue.deprecationReason")
     ),
     Test / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oF"),
     libraryDependencies ++= Seq(
--- a/modules/core/src/main/scala/sangria/introspection/model.scala
+++ b/modules/core/src/main/scala/sangria/introspection/model.scala
@@ -115,7 +115,10 @@ case class IntrospectionInputValue(
     name: String,
     description: Option[String],
     tpe: IntrospectionTypeRef,
-    defaultValue: Option[String])
+    defaultValue: Option[String],
+    isDeprecated: Option[Boolean],
+    deprecationReason: Option[String]
+)
 
 sealed trait IntrospectionTypeRef {
   def kind: TypeKind.Value
--- a/modules/core/src/main/scala/sangria/schema/Schema.scala
+++ b/modules/core/src/main/scala/sangria/schema/Schema.scala
@@ -708,6 +708,7 @@ trait InputValue[T] {
   def inputValueType: InputType[_]
   def description: Option[String]
   def defaultValue: Option[(_, ToInput[_, _])]
+  def deprecationReason: Option[String]
 }
 
 /** @param description
@@ -719,17 +720,22 @@ case class Argument[T](
     description: Option[String],
     defaultValue: Option[(_, ToInput[_, _])],
     fromInput: FromInput[_],
+    deprecationReason: Option[String],
     astDirectives: Vector[ast.Directive],
     astNodes: Vector[ast.AstNode])
     extends InputValue[T]
     with Named
-    with HasAstInfo {
+    with HasAstInfo
+    with HasDeprecation {
 
   def withDirective(directive: ast.Directive): Argument[T] =
     copy(astDirectives = astDirectives :+ directive)
 
   def withDirectives(directives: ast.Directive*): Argument[T] =
     copy(astDirectives = astDirectives ++ directives)
+
+  def withDeprecationReason(deprecationReason: String): Argument[T] =
+    copy(deprecationReason = Some(deprecationReason))
 
   override def inputValueType: InputType[_] = argumentType
   override def rename(newName: String): this.type = copy(name = newName).asInstanceOf[this.type]
@@ -751,6 +757,7 @@ object Argument {
       Some(description),
       Some(defaultValue -> toInput),
       fromInput,
+      None,
       Vector.empty,
       Vector.empty)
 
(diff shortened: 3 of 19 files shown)


yanns:
Related to https://github.com/sangria-graphql/sangria/pull/992

asteinwedel:
(also just updated something so working on tests)

> Related to #992

hahah should have looked first 🤦‍♀️

asteinwedel:
@yanns an issue I have with https://github.com/sangria-graphql/sangria/pull/992 is it requires `isDeprecated` on `IntrospectionInputValue` and automatically adds it to the introspection query (can see in `graphql-js` that these are optional)

otherwise, I think we've mainly covered the same things though I've touched more places with checking deprecation

yanns:
Yes I have the same issue with the introspection query.
I was thinking about adding a flag so that the user can decide whether to use it or not, but I never put time to finish this :D

asteinwedel:
alright I added deprecation tracking and directive testing in various places

there is a failing test in `IntrospectionSchemaMaterializerSpec` in `testSchema` where `serverSchema` != `clientSchema`, but I'll have to check that out later

asteinwedel:
@yanns I have tests and everything working now, but I'm a bit unfamiliar with how versioning works here

it looks like we're testing against sangria-core 2.13.12 and failing some checks

a few different things in this PR would be breaking changes, requiring a major version update, not sure what the process like that is here

yanns:
First you can add the breaking changes in the build.sbt to that the build can run.
Depending on the number of breaking changes, I will decide for a major release or not.
Thanks for all your efforts!!!

yanns:
There's one issue on Scala 3: https://github.com/sangria-graphql/sangria/actions/runs/6471747880/job/17571053945?pr=1063
strange

asteinwedel:
> There's one issue on Scala 3: https://github.com/sangria-graphql/sangria/actions/runs/6471747880/job/17571053945?pr=1063 strange

hm i had a check to prevent this infinite recursion

I also noticed I'm missing something with nested input objects

so still looking at these things

q: what about splitting out deprecation tracking to a different PR? would it be okay to add support minus `DeprecationTracker` work? since that's a significantly larger change

yanns:
Yes I wanted to ask why you have added deprecation tracking in this PR.
Isn't it something we can achieve with a middleware?
In general, yes to split this. 💯

asteinwedel:
> Yes I wanted to ask why you have added deprecation tracking in this PR.
> Isn't it something we can achieve with a middleware?
> In general, yes to split this. 💯

no we aren't actually using the deprecation tracker so it's not as big of a deal for my purpose, I care more about supporting querying the introspection schema

I was adding it just to be consistent with other deprecation parts, but it's become quite big 😅 I'll split that out

modules/core/src/main/scala/sangria/introspection/model.scala
```diff
@@ -137,4 +140,5 @@ case class IntrospectionDirective(
     description: Option[String],
     locations: Set[DirectiveLocation.Value],
     args: Seq[IntrospectionInputValue],
-    repeatable: Boolean)
+    repeatable: Boolean
+)
```
yanns, line 144: nitpick: this change is not necessary

modules/core/src/main/scala/sangria/schema/Schema.scala
```diff
@@ -720,17 +721,22 @@ case class Argument[T](
     defaultValue: Option[(_, ToInput[_, _])],
     fromInput: FromInput[_],
     astDirectives: Vector[ast.Directive],
-    astNodes: Vector[ast.AstNode])
+    astNodes: Vector[ast.AstNode],
+    deprecationReason: Option[String])
```
yanns, line 725: `Field` and `EnumValue` defines the fields in this order:
- `deprecationReason: Option[String]`
- `astDirectives: Vector[ast.Directive]`
- `astNodes: Vector[ast.AstNode]`

It'd be great to keep the same order for `InputField` and `Argument`.

modules/core/src/main/scala/sangria/schema/Schema.scala
```diff
@@ -741,18 +747,21 @@ object Argument {
       name: String,
       argumentType: InputType[T],
       description: String,
-      defaultValue: Default)(implicit
+      defaultValue: Default
+  )(implicit
```
yanns, line 751: Nitpick: this change is not necessary

modules/core/src/main/scala/sangria/schema/Schema.scala
```diff
@@ -741,18 +747,21 @@ object Argument {
...
       toInput: ToInput[Default, _],
       fromInput: FromInput[T],
-      res: ArgumentType[T]): Argument[res.Res] =
+      res: ArgumentType[T]
+  ): Argument[res.Res] =
```
yanns, line 755: Nitpick: this change is not necessary

modules/core/src/test/scala/sangria/introspection/IntrospectionSpec.scala
```diff
@@ -536,7 +562,8 @@ class IntrospectionSpec extends AnyWordSpec with Matchers with FutureResultSuppo
                     "name" -> "includeDeprecated",
                     "description" -> null,
                     "type" -> Map("kind" -> "SCALAR", "name" -> "Boolean", "ofType" -> null),
-                    "defaultValue" -> "false")),
+                    "defaultValue" -> "false"
+                  )),
```
yanns, line 566: Nitpick: this change is not necessary

modules/core/src/test/scala/sangria/introspection/IntrospectionSpec.scala
```diff
@@ -582,7 +609,8 @@ class IntrospectionSpec extends AnyWordSpec with Matchers with FutureResultSuppo
...
```
yanns, line 613: Nitpick: this change is not necessary

modules/core/src/test/scala/sangria/schema/SchemaComparatorSpec.scala
```diff
@@ -203,13 +204,18 @@ class SchemaComparatorSpec extends AnyWordSpec with Matchers {
...
       """,
       nonBreakingChange[TypeAdded]("`Int` type was added"),
       breakingChange[InputFieldRemoved]("Input field `descr` was removed from `Filter` type"),
       nonBreakingChange[InputFieldAdded]("Input field `size` was added to `Filter` type"),
       nonBreakingChange[InputFieldDescriptionChanged]("`Filter.name` description is changed"),
-      nonBreakingChange[TypeDescriptionChanged]("`Filter` type description is changed")
+      nonBreakingChange[TypeDescriptionChanged]("`Filter` type description is changed"),
+      nonBreakingChange[InputFieldDeprecated]("`Filter.foo` was deprecated"),
+      nonBreakingChange[InputFieldAstDirectiveAdded](
+        "Directive `@deprecated` added on an input field `Filter.foo`")
```
yanns, line 218: 👍

modules/core/src/test/scala/sangria/schema/SchemaComparatorSpec.scala
```diff
@@ -297,7 +305,11 @@ class SchemaComparatorSpec extends AnyWordSpec with Matchers {
       nonBreakingChange[ObjectTypeArgumentDefaultChanged](
         "`Filter.foo(a)` default value changed from none to `\"foo\"`"),
       nonBreakingChange[ObjectTypeArgumentDescriptionChanged](
-        "`Filter.foo(a)` description is changed")
+        "`Filter.foo(a)` description is changed"),
+      nonBreakingChange[ObjectTypeArgumentDeprecated](
+        "Argument `c1` on `Filter.foo` was deprecated"),
+      nonBreakingChange[FieldArgumentAstDirectiveAdded](
+        "Directive `@deprecated` added on a field argument `Filter.foo[c1]`")
```
yanns, line 312: 👍

modules/core/src/test/scala/sangria/schema/SchemaComparatorSpec.scala
```diff
@@ -482,7 +494,10 @@ class SchemaComparatorSpec extends AnyWordSpec with Matchers {
       nonBreakingChange[DirectiveArgumentAstDirectiveRemoved](
         "Directive `@hello` removed from a directive argument `test.bar`"),
       nonBreakingChange[DirectiveArgumentAstDirectiveAdded](
-        "Directive `@world` added on a directive argument `test.bar`")
+        "Directive `@world` added on a directive argument `test.bar`"),
+      nonBreakingChange[DirectiveArgumentAstDirectiveAdded](
+        "Directive `@deprecated` added on a directive argument `test.foo`"),
+      nonBreakingChange[DirectiveArgumentDeprecated]("Directive argument `test.foo` was deprecated")
```
yanns, line 500: 👍

modules/derivation/src/main/scala/sangria/macros/derive/DeriveInputObjectSetting.scala
```diff
@@ -9,6 +9,7 @@ case class InputObjectTypeDescription(description: String) extends DeriveInputOb
 
 case class DocumentInputField(fieldName: String, description: String)
     extends DeriveInputObjectSetting
+
```
yanns, line 12: Nitpick: this change is not necessary

yanns (commented):
This is looking very good! Congrats for all new tests!

I left some minor comments to minimize the git diff.
The "real" comment is about the order of the fields.

asteinwedel:
updated!

build.sbt
```diff
@@ -211,7 +238,7 @@ lazy val sangriaCatsEffectExperimental = project
     name := "sangria-cats-effect-experimental",
     description := "Experimental support for Cats Effect",
     libraryDependencies ++= List(
-      "org.typelevel" %% "cats-effect" % "3.5.1",
+      "org.typelevel" %% "cats-effect" % "3.5.2",
```
yanns, line 245: Nitpick: this change is not related to the PR.

project/plugins.sbt
```diff
@@ -2,6 +2,6 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "1.1.3")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.9")
 addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.3.11")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.6")
-addSbtPlugin("com.github.sbt" % "sbt-github-actions" % "0.16.0")
+addSbtPlugin("com.github.sbt" % "sbt-github-actions" % "0.17.0")
```
yanns, line 5: Nitpick: this change is not related to the PR.

yanns (commented):
It's looking really good!
Question: why do you have picked 2 commits unrelated to this PR? (the dependency updates)

asteinwedel:
ah sorry that was a rebase mistake

yanns (approved):
🥇 awesome, thank you for your contribution!

asteinwedel:
@yanns could we get a release for this particularly? need to support introspection queries with this, could be a dev prerelease

yanns:
You can try this pre-release: https://github.com/sangria-graphql/sangria/releases/tag/v4.1.0-RC1
It can take a few hours to synchronize with maven central.

asteinwedel:
> You can try this pre-release: https://github.com/sangria-graphql/sangria/releases/tag/v4.1.0-RC1 It can take a few hours to synchronize with maven central.

thanks! works :)

PR #1064, opened 2023-10-12
Input deprecation tracker

asteinwedel:
I have a question about accessing arg values, will comment below


modules/core/src/main/scala/sangria/execution/FutureResolver.scala
```diff
@@ -1534,6 +1539,140 @@ private[execution] class FutureResolver[Ctx](
...
+    }
+
+    def trackDeprecatedInputObjectFields(inputType: InputType[_], ioArg: Input): Unit =
+      inputType match {
+        case ioDef: InputObjectType[_] =>
+          ioDef.fields.foreach { field =>
+            // field deprecation
+            val fieldVal: Option[Input] = (ioArg match {
+              case lm: ListMap[String, Input] => lm.get(field.name)
+              case in: Input =>
```
asteinwedel, line 1601: I'm a little confused about how `Args` is set at this point

at first I was assuming it was all `Input` and didn't have the `ListMap` check here (and didn't have the `Seq` check below for lists) but that was causing errors

it doesn't appear to actually be `Input` although everything compiles fine, I had to add a `try { } catch { }` for `None` values being here in some tests

removing the input unmarshaller and `Input` bit entirely seems to actually still work just fine, so think I can just remove that? (going to double check and make a new commit for this)

asteinwedel:
@ https://github.com/sangria-graphql/sangria/pull/1064/files#r1356792135

latest commit removed the input unmarshalling and everything still seems to work

modules/core/src/main/scala/sangria/execution/FutureResolver.scala
```diff
@@ -1534,6 +1535,124 @@ private[execution] class FutureResolver[Ctx](
...
+      }
+
+    def deprecatedArgsUsed(argDefs: List[Argument[_]], argValues: Args): List[Argument[_]] =
+      argDefs.filter { argDef =>
+        val argValue = getArgValue(argDef.name, argValues)
+        argDef.deprecationReason.isDefined && argValue.isDefined
+      }
+
+    def trackDeprecatedDirectiveArgs(astDirective: ast.Directive): Unit = {
+      // prevent infinite loop from directiveA -> arg -> directiveA -> arg ...
```
asteinwedel, line 1560: I'm not sure if this is something we can actually run into with query parsing, but it is something that was possible to run into with tests

modules/core/src/main/scala/sangria/execution/Resolver.scala
```diff
@@ -27,10 +27,7 @@ private[execution] trait ResolverBuilder {
       preserveOriginalErrors: Boolean,
       validationTiming: TimeMeasurement,
       queryReducerTiming: TimeMeasurement,
-      queryAst: ast.Document
-  )(implicit
-      executionContext: ExecutionContext
-  ): Resolver[Ctx]
+      queryAst: ast.Document)(implicit executionContext: ExecutionContext): Resolver[Ctx]
```
yanns, line 30: Nitpick: this change is not necessary.

modules/core/src/main/scala/sangria/execution/FutureResolver.scala
```diff
@@ -1572,8 +1691,7 @@ private[execution] class FutureResolver[Ctx](
               deferredResolverState = deferredResolverState
             )
 
-            if (allFields.exists(_.deprecationReason.isDefined))
-              deprecationTracker.deprecatedFieldUsed(ctx)
+            trackDeprecation(ctx)
```
yanns, line 1694: The deprecation logic seems more expensive now.
If we're using `DeprecationTracker.empty`, we could actually avoid this work.
What do you think about making `DeprecationTracker` optional in `Executor` (as we're breaking changes anyway), and avoid doing all this work is it's `None`?

modules/core/src/main/scala/sangria/execution/FutureResolver.scala
```diff
@@ -1534,6 +1535,124 @@ private[execution] class FutureResolver[Ctx](
...
+
+    def deprecatedArgsUsed(argDefs: List[Argument[_]], argValues: Args): List[Argument[_]] =
+      argDefs.filter { argDef =>
+        val argValue = getArgValue(argDef.name, argValues)
+        argDef.deprecationReason.isDefined && argValue.isDefined
+      }
+
+    def trackDeprecatedDirectiveArgs(astDirective: ast.Directive): Unit = {
+      // prevent infinite loop from directiveA -> arg -> directiveA -> arg ...
+      if (visitedDirectives.contains(astDirective)) {
```
yanns, line 1561: Could we  save and check only the directive name to have a more efficient `Set`?

modules/core/src/main/scala/sangria/execution/FutureResolver.scala
asteinwedel: good idea!

asteinwedel:
I went ahead and entirely removed the empty deprecation tracker, it didn't seem to serve any purpose now that it's optional and leaving it could have someone unintentionally going through all the tracker work when it's not needed so thought being clear about that would be better

modules/core/src/main/scala/sangria/execution/FutureResolver.scala
```diff
@@ -1572,8 +1693,7 @@ private[execution] class FutureResolver[Ctx](
               deferredResolverState = deferredResolverState
             )
 
-            if (allFields.exists(_.deprecationReason.isDefined))
-              deprecationTracker.deprecatedFieldUsed(ctx)
+            deprecationTracker.map(trackDeprecation(_, ctx))
```
yanns, line 1696: We're returning `Unit`, so I think no need to map it.
```suggestion
            deprecationTracker.foreach(trackDeprecation(_, ctx))
```

modules/core/src/main/scala/sangria/execution/ValueCoercionHelper.scala
```diff
@@ -527,7 +527,7 @@ class ValueCoercionHelper[Ctx](
                   isArgument))),
           { case (v, deprecated) =>
             if (deprecated && userContext.isDefined)
-              deprecationTracker.deprecatedEnumValueUsed(enumT, v, userContext.get)
+              deprecationTracker.map(_.deprecatedEnumValueUsed(enumT, v, userContext.get))
```
yanns, line 530:
```suggestion
              deprecationTracker.foreach(_.deprecatedEnumValueUsed(enumT, v, userContext.get))
```

yanns (commented):
It's looking good.
Do you think there's some code I should really focus on?

modules/core/src/main/scala/sangria/execution/FutureResolver.scala
asteinwedel: habit from mostly working with IO stuff 😅 updating

asteinwedel:
> Do you think there's some code I should really focus on?
sorry for the late response, had a few other things on my plate last week

`trackDeprecation` is the bulk of it, and tried to cover it thoroughly with the unit tests. I think the main parts are

1. track directive args (recursively since args can also have directives on them)
2. track field args (at least this was simple) + their directives+args
3. track input objects (recursively since fields could be input objects) + their directives+args

yanns (approved):
👍 it's looking good, thanks for the tests!

PR #1067, opened 2023-10-17
include enums on directive args in schema introspection

asteinwedel:
Currently we're only including unused types in the additional types of ast based schemas. This causes enums that are only used on directive args to be missed when generating an introspection query result, as we are not collecting types on directive args.


yanns (approved):
💯

yanns:
Do you need a quick release for this change? Or do you have other changes on view before?

asteinwedel:
> Do you need a quick release for this change? Or do you have other changes on view before?

would very appreciate a release with the other changes I made, thanks!

Issue #1068, opened 2023-10-20
Support OneOf on Input Objects


Tehnix:
It seems `oneOf` has largely been accepted, and many GraphQL libraries have either started implementation or already rolled it out https://github.com/graphql/graphql-spec/pull/825.

Is there any interest in Sangria supporting this, and would there be any technical blockers for it?

At $work we have had several cases where support for this would have made the resulting API better and more intuitive.

For inspiration, there's ongoing work in caliban here https://github.com/ghostdogpr/caliban/pull/1846.

yanns:
Yes, this can be interesting.
I don't think I'll have time to work on it in the next weeks though.

yanns:
released in https://github.com/sangria-graphql/sangria/releases/tag/v4.2.0

PR #1119, opened 2024-07-18
Support oneOf directive on input objects

asteinwedel:
Issue:
https://github.com/sangria-graphql/sangria/issues/1068

WIP.

The main thing I'm missing is when checking that there's "exactly one non-null field" on a query (I have the `ExactlyOneOfFieldGiven` rule), I'm struggling with validating against `ast.VariableValue` (since I need the unmarshalled input for that). (and considering default values for `VariableDefinition` -- though that seems easier than getting the actual input)

was wondering if it needs to go in the execution part for checking variables, but that seems a bit messier.

Trying to figure it out...but feel free to contribute if you have a way to go about that

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/Executor.scala
+++ b/modules/core/src/main/scala/sangria/execution/Executor.scala
@@ -29,107 +29,125 @@ case class Executor[Ctx, Root](
       operationName: Option[String] = None,
       variables: Input = emptyMapVars
   )(implicit um: InputUnmarshaller[Input]): Future[PreparedQuery[Ctx, Root, Input]] = {
-    val (violations, validationTiming) =
-      TimeMeasurement.measure(queryValidator.validateQuery(schema, queryAst, errorsLimit))
+    val scalarMiddleware = Middleware.composeFromScalarMiddleware(middleware, userContext)
+    val valueCollector = new ValueCollector[Ctx, Input](
+      schema,
+      variables,
+      queryAst.sourceMapper,
+      deprecationTracker,
+      userContext,
+      exceptionHandler,
+      scalarMiddleware,
+      false)(um)
 
-    if (violations.nonEmpty)
-      Future.failed(ValidationError(violations, exceptionHandler))
-    else {
-      val scalarMiddleware = Middleware.composeFromScalarMiddleware(middleware, userContext)
-      val valueCollector = new ValueCollector[Ctx, Input](
-        schema,
-        variables,
-        queryAst.sourceMapper,
-        deprecationTracker,
-        userContext,
-        exceptionHandler,
+    val operationCtx = for {
+      operation <- Executor.getOperation(exceptionHandler, queryAst, operationName)
+      unmarshalledVariables <- valueCollector.getVariableValues(
+        operation.variables,
         scalarMiddleware,
-        false)(um)
+        errorsLimit
+      )
+    } yield (operation, unmarshalledVariables)
 
-      val executionResult = for {
-        operation <- Executor.getOperation(exceptionHandler, queryAst, operationName)
-        unmarshalledVariables <- valueCollector.getVariableValues(
-          operation.variables,
-          scalarMiddleware,
-          errorsLimit
-        )
-        fieldCollector = new FieldCollector[Ctx, Root](
-          schema,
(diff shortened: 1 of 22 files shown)


yanns:
first of all, thanks for your contribution!
I see it's already going well, with good test coverage.

For the missing validation, do you already have failing tests for it? It would save me time.

asteinwedel:
> first of all, thanks for your contribution! I see it's already going well, with good test coverage.
> 
> For the missing validation, do you already have failing tests for it? It would save me time.

I do not, as the tests I have setup aren't setup for variables.

Testing something like this would be a basic case (and then also default values)
```
val schema = gql"""
   type Query {
      oneOfQuery(input: OneOfInput!)
  }
  
  input OneOfInput @oneOf {
    foo: String
    bar: Int
  }
"""
val query = """
  query Test($foo: String) {
    oneOfQuery(input: { foo: $foo, bar: 123)
  }

val variables = """{ "foo": "testing" }"""
"""

File: modules/core/src/main/scala/sangria/execution/Executor.scala [hunk 1]
```diff
@@ -29,107 +29,116 @@ case class Executor[Ctx, Root](
       operationName: Option[String] = None,
       variables: Input = emptyMapVars
   )(implicit um: InputUnmarshaller[Input]): Future[PreparedQuery[Ctx, Root, Input]] = {
-    val (violations, validationTiming) =
```
asteinwedel, line 32: this is mostly just moving things around so `queryValidator.validateQuery` has access to `unmarshalledVariables`

asteinwedel:
@yanns I got something working, though it causes some breaking changes with `QueryValidator` needing variable values now.

working on making sure build /test passes

but that change is in its own commit

asteinwedel:
hmm moving validation after variable coercion might need a bit more tuning to make sure validation errors get thrown before variable coercion errors

asteinwedel:
alright...seems passing now

asteinwedel:
let me know if think of a better way to check the variable values

yanns:
Just taking notes:
- we need to validate: "making a OneOf Input Object into a regular Input Object is a non-breaking change (the reverse is a breaking change)"
- would it be better to follow [all test cases documented in the spec](https://github.com/graphql/graphql-spec/pull/825/files#diff-30a69c5a5eded8e1aea52e53dad1181e6ec8f549ca2c50570b035153e2de1c43R1728-R1766)?
- about the validation of variables with the `QueryValidator`: it's indeed quite a big change. But I don't see how we can do without this. We could split the validation in 2: one without and one with the variables. But I found it's quite nice to have all validations inside `ExactlyOneOfFieldGiven`. Maybe I could release a release candidate version, and we could check if we're breaking much.

yanns (approved):
looking good to me. I'll release a RC to test that with more code.

yanns:
I almost forget: we need to test for breaking changes (https://github.com/sangria-graphql/sangria/blob/5f775794a74037c19b27cb03125be5f8f2196818/modules/core/src/main/scala/sangria/schema/SchemaComparator.scala#L603)

asteinwedel:
ah thanks for the spec link!

looks like the official spec doesn't like `null` values either which is the main difference, updated tests and made sure those cases were all covered

(from spec: I didn't do type checking since that's another area of sangria validation)

will work on the breaking change one now

yanns:
yes sorry, it was easier for me to push some changes directly. I've also pushed the changes for the breaking changes.

yanns:
yes sorry, it was easier for me to push some changes directly. I've also pushed the changes for the breaking changes.

asteinwedel:
> yes sorry, it was easier for me to push some changes directly. I've also pushed the changes for the breaking changes.

got it I pushed the null check change now

yanns:
oh miss, I force pushed some changes as the last commit from me was not complete (was missing the exceptions for the mina checks).
Now I'm over.
Can you push your changes again?
Edit: i see you did, thanks!

yanns (approved):
Is there anything left you'd like to add / change?

asteinwedel:
> Is there anything left you'd like to add / change?

nope think this covers it all! thanks for the help!

PR #1129, opened 2024-08-22
Fix: Schema rendering problems with Windows line separators

### LINKED ISSUE #1125: Schema rendering problems with Windows line separators
If source files are using Windows line separators, then the schema renderer seems to include the `\r\n` line breaks inside multi-line comments from the source files. The renderer logic seems to expect `\n` line breaks though, and because of that creates lines that contain only spaces (from the indentation).

Example of unexpected output, using `+` instead of space as indentation for readability:

```
type SomeType {
++"""
++Some value.
++
++The line above should be empty.
++"""
++value: String!
}
```

It seems like the cause of the problem is just this single statement:

https://github.com/sangria-graphql/sangria/blob/595999522f4915069709ca0499524959d81ecf70/modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala#L480-L483

Changing `.split("\n").iterator` to `.linesIterator` should probably fix the issue.

ArminWiebigke:
Fix #1125

### DIFF
--- a/modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala
+++ b/modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala
@@ -477,7 +477,7 @@ object QueryRenderer {
       extraIndent: Boolean = true): String =
     if (node.value.trim.nonEmpty) {
       val ind = if (extraIndent) indent.incForce.str else indent.strForce
-      val lines = escapeBlockString(node.value).split("\n").iterator.map { line =>
+      val lines = escapeBlockString(node.value).linesIterator.map { line =>
         if (line.isEmpty) line // do not output lines with only whitespaces inside
         else ind + line
       }
--- a/modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala
+++ b/modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala
@@ -1,8 +1,7 @@
 package sangria.renderer
 
-import sangria.ast.AstLocation
-import sangria.util.StringUtil.{escapeBlockString, escapeString}
 import sangria.ast._
+import sangria.util.StringUtil.{escapeBlockString, escapeString, linesIterator}
 
 object QueryRenderer {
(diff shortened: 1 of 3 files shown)


yanns:
Thanks a lot! I very like the test case you added!

The CI checks are failing with a compilation error:
```
[error] /home/runner/work/sangria/sangria/modules/core/src/main/scala/sangria/renderer/QueryRenderer.scala:480:63: value iterator is not a member of Iterator[String]
[error]       val lines = escapeBlockString(node.value).linesIterator.iterator.map { line =>
[error]     
```
Do you want to take care of this?

yanns:
One test seem to fail:
```
[info]   - should renders comments with various line separators correctly *** FAILED ***
[info]     "..._LF:
[info]     
[info]     __CRLF:
[info]     
[info]     __CR:[

]Empty_lines_above_sh..." did not equal "..._LF:
[info]     
[info]     __CRLF:
[info]     
[info]     __CR:[
[info]     
[info]     __]Empty_lines_above_sh..." (QueryRendererSpec.scala:1018)
[info]     Analysis:
[info]     "..._LF:
[info] 
[info] __CRLF:
[info] 
[info] __CR:[
```

ArminWiebigke:
Seems like `\r` is not considered a valid line break in Scala 2.12.19 🤔
So I guess we just need to split manually in case any `\r`s are still contained in the lines returned by `linesIterator`.

yanns:
> Seems like `\r` is not considered a valid line break in Scala 2.12.19 🤔 So I guess we just need to split manually in case any `\r`s are still contained in the lines returned by `linesIterator`.

Having some code just for scala 2.12 is not optimal.
Why if we would copy / paste the implementation of `linesIterator` into our code (I mean the implementation that is also considering `\r`)?

yanns (approved):
💯

yanns:
I was checking the specs, and your code seems to be compliant with it: https://spec.graphql.org/October2021/#LineTerminator

PR #1144, opened 2024-10-09
Fix enum derivation

wordrak:
Fix enum derivation when the enum composed of case objects but they are not within another class

### DIFF
--- a/modules/derivation/src/main/scala-2/sangria/macros/derive/DeriveEnumTypeMacro.scala
+++ b/modules/derivation/src/main/scala-2/sangria/macros/derive/DeriveEnumTypeMacro.scala
@@ -132,7 +132,7 @@ class DeriveEnumTypeMacro(context: blackbox.Context)
 
         val actualValue =
           if (value.isModuleClass) {
-            if (value.owner.isModuleClass) {
+            if (value.owner.isModuleClass && !value.owner.isPackage) {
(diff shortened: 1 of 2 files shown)


yanns:
@wordrak thanks a lot for the PR.
To help me reviewing this, could you explain more what the actual issue is?

wordrak:
Hi, issue is that the `deriveEnumType` macro fails when the Enum is a "case object hierarchy" and it is not contained in another object. 

e.g.:

```
  sealed trait Fruit

  case object RedApple extends Fruit
  case object SuperBanana extends Fruit
  case object MegaOrange extends Fruit
  
  object Fruit {
    implicit val enumType = deriveEnumType[Fruit]()
  }
```

If you revert the changes in the TestCase then you see the problem.

yanns:
instead of creating the term `derive.SuperBanana` that is failing:
```
[error] /Users/yannsimon/projects/sangria-graphql/sangria/modules/derivation/src/test/scala/sangria/macros/derive/DeriveEnumTypeMacroSpec.scala:55:48: not found: value derive
[error]       val singletonEnum = deriveEnumType[Fruit]()
[error]                                                ^
```
we know create the term `SuperBanana`.

yanns (approved):

yanns:
Thanks a lot!

Issue #1041, opened 2023-08-11
Unnecessary implicit conversion while using experimental catseffect `AsyncValue`


roman0x58:
Hello there, I'm trying to eliminate the need for the Future with the functionality described here: https://github.com/sangria-graphql/sangria/pull/987. However, I'm encountering a lot of ambiguity errors in places where F[X] is used. The implicit conversion in the AsyncValue object is attempting to convert the types to LeafAction. Please see the errors above for more details. Do you have any advice on how to address this ambiguity? Thanks!

1. 
```

[error] 133 |        parentPrice <- fs2.Stream.eval(collection(ctx).findById(parentId).compile.toList.map(_.headOption))
[error]     |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |Found:    F[List[O2]]
[error]     |Required: ?{ map: ? }
[error]     |
[error]     |where:    F  is a type in class PriceDictionary with bounds <: [_] =>> Any
[error]     |          O2 is a type variable with constraint >: model.TreeElement[model.PriceItem]
[error]     |
[error]     |Note that implicit extension methods cannot be applied because they are ambiguous;
[error]     |both method asyncAction in object AsyncValue and method toFunctorOps in trait ToFunctorOps provide an extension method `map` on (fs2.Stream.CompileOps[[x] =>> F[x], [x] =>> F[x], O2]#toList : => F[List[O2]])
```

2.
 ```
[error] 86 |    (for {
[error] 87 |      user <- c.ctx.om.mongo.users
[error] 88 |        .findOne(or(equal("email", email), equal("login", email)))
[error] 89 |        .compile
[error] 90 |        .toList
[error] 91 |        .map(_.headOption)
[error] 92 |    } yield user match {
[error] 93 |      case Some(u) => checkAuthContext(u, c.arg(Context))(process(_))
[error] 94 |      case _ => failForm[UserPassport]("email", messages("om.auth.violation.notfound"))
[error] 95 |    }).flatten
[error]    |    ^
[error]    |value flatten is not a member of sangria.schema.LeafAction[Nothing, F[auth.UserPassport]], but could be made available as an extension method.
```

yanns:
Thanks for the report.
Indeed, the scala compiler seems to transform the `F[_]` into a `LeafAction` too early.
Maybe I'll need to push the implicit conversion into a trait to give it less priority.
I can try that later.

[...]

PR #1156, opened 2024-11-07
try to reproduce issue #1041

yanns:


roman0x58:
@yanns I think there should be something like this, it should reproduce the issue:
```scala
object IOExecutionSchemeSpec {
  import sangria.catseffect.schema.AsyncValue._

  private def resolve[F[_]: Async](): F[String] = {
    import cats.syntax.functor._
    Async[F].pure(Option("hello")).map {
      case Some(value) => value
      case None => throw new Exception("No value")
    }
  }

  private val QueryType: ObjectType[Unit, Unit] = ObjectType(
    "Query",
    () =>
      fields[Unit, Unit](
        Field("ids", ListType(IntType), resolve = _ => List(1, 2)),
        Field(
          "parent",
          StringType,
          resolve = { _ => resolve[IO]() }
        )
      )
  )

  private val schema = Schema(QueryType)
}
````

**Comment by yanns** [comment]
Thanks, I can now reproduce the issue.
The solution I had by using low priority implicits sadly does not work.
I don't have any solution for this.

**Comment by roman0x58** [comment]
@yanns I played around with this issue a bit, maybe we could use something like this:
```scala
  def async[Ctx, Val, Res, Out, F[_]](
                                 name: String,
                                 fieldType: OutputType[Out],
                                 description: Option[String] = None,
                                 arguments: List[Argument[_]] = Nil,
                                 resolve: Context[Ctx, Val] => F[Res],
                                 possibleTypes: => List[PossibleObject[_, _]] = Nil,
                                 tags: List[FieldTag] = Nil,
                                 complexity: Option[(Ctx, Args, Double) => Double] = None,
                                 deprecationReason: Option[String] = None,
                                 astDirectives: Vector[ast.Directive] = Vector.empty
                               )(implicit ev: ValidOutType[Res, Out],
                                 asyncToAction: (Context[Ctx, Val] => F[Res]) => Context[Ctx, Val] => Action[Ctx, Res]): Field[Ctx, Val] =
    Field[Ctx, Val](
      name,
      fieldType,
      description,
      arguments,
      asyncToAction(resolve),
      deprecationReason,
      tags,
      complexity,
      () => possibleTypes.map(_.objectType),
      astDirectives,
      Vector.empty)
```
and the resolver implicit conversion: 
```scala
object AsyncResolver {
  implicit def asyncToAction[Ctx, Val, Res, F[_]: Async](resolver: Context[Ctx, Val] => F[Res]): Context[Ctx, Val] => Action[Ctx, Res] = {
    context => AsyncValue(resolver(context))
  }
}
```
This will require an additional signature for Field, but we’ll avoid the ambiguity issue. Or alternatively, we could simply expose the `asyncToAction` function to wrap the `resolve` function explicitly.

Issue #1162, opened 2024-12-13
LongType doesn't support incoming String values


jakobmerrild:
Under the assumption that most consumers of the GraphQL schema specified by Sangria is running in JavaScript it seems prudent to allow input coercion from `String` for the `Long` scalar type, as JavaScript has no way of precisely representing the full spectrum using it's `number` data type.

I.e. given the following schema
```graphql
type Query {
  echo(number: Long!): Long!
}
```
then the following payload should be accepted
```
{
  "query": "query EchoLong($number: Long!) { echo(number: $number) }",
  "operationName": "EchoLong",
  "variables": { "number": "123456" }
}
```

yanns:
Thanks for the issue and the PR!
To help me make a decision, do you know if the GraphQL specs mention this?
Are you also aware of how other libraries are handling this?

jakobmerrild:
According to the latest spec services that specify custom scalars should include the `@specifiedBy` url, see https://spec.graphql.org/October2021/#sec-Scalars.Custom-Scalars

However, I don't know if Sangria currently supports that annotation?

W.r.t coercion it is up to the implementor to define what are reasonable rules so long as they are performed without losing precision (if I'm reading the [spec](https://spec.graphql.org/October2021/#sec-Scalars.Result-Coercion-and-Serialization) correctly)

jakobmerrild:
According to this [announcement](https://graphql.org/blog/2023-01-14-graphql-scalars/) you can contribute your custom scalars, though given that the announcement is almost 2 years old and there are [no new contributions](https://scalars.graphql.org/) by the community since then, it might not really have taken hold as an idea

yanns:
Your issue is about input coercion right? https://spec.graphql.org/October2021/#sec-Scalars.Input-Coercion

What you mentioned before is not the same:
> Since this coercion behavior is **not observable to clients of the GraphQL service**, the precise rules of coercion are left to the implementation. 

This is for example if the field is a `String`, and internally it's an Integer, then "the precise rules of coercion are left to the implementation. ".


I assume you're having this issue with a real-world case. Out of curiosity, can you provide more details, like which JavaScript library is used?

yanns:
Input coercion for the scalar `Int` do not accept `String` values:
https://spec.graphql.org/October2021/#sec-Int.Input-Coercion
> When expected as an input type, only integer input values are accepted. All other input values, including strings with numeric content, must raise a request error indicating an incorrect type

[...]

PR #1163, opened 2024-12-13
Allow string inputs for Long Scalar type

jakobmerrild:
Solves: https://github.com/sangria-graphql/sangria/issues/1162

Let me know if this is really something that should be solved in `sangria-circe` instead?

### DIFF
--- a/modules/core/src/main/scala/sangria/schema/package.scala
+++ b/modules/core/src/main/scala/sangria/schema/package.scala
@@ -2,6 +2,7 @@ package sangria
 
 import sangria.marshalling.MarshallerCapability
 import sangria.validation._
+import scala.util.Try
 
 /** Types that describe a GraphQL schema.
   *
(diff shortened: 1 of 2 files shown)


yanns:
There's a compilation error with scala 2.12:
```
 [error] /home/runner/work/sangria/sangria/modules/core/src/main/scala/sangria/schema/package.scala:50:27: value toLongOption is not a member of String
[error] Error occurred in an application involving default arguments.
[error]       case s: String if s.toLongOption.isDefined => Right(s.toLong)
[error]                           ^
[error] /home/runner/work/sangria/sangria/modules/core/src/main/scala/sangria/schema/package.scala:57:48: value toLongOption is not a member of String
[error] Error occurred in an application involving default arguments.
[error]       case ast.StringValue(s, _, _, _, _) if s.toLongOption.isDefined => Right(s.toLong)
[error]                                                ^
```

jakobmerrild:
I tried to create a specification for a custom `Long` scalar: https://github.com/graphql/graphql-scalars/pull/26

Issue #906, opened 2022-09-05
Scala 3: Exception returning Future values from GraphQL Endpoint


PsyfireX:
When attempting to return values from a GraphQL endpoint using Sangria and Scala 3, I get the following exception:

```
    @GraphQLField
    def ping(in: String): Future[String] = {
      implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global
      Future.apply(in)
    }
```

```
"class scala.concurrent.impl.Promise$Transformation cannot be cast to class java.lang.String (scala.concurrent.impl.Promise$Transformation is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap') 
 java.lang.ClassCastException: class scala.concurrent.impl.Promise$Transformation cannot be cast to class java.lang.String (scala.concurrent.impl.Promise$Transformation is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap') 
 sangria.execution.Resolver.resolveValue(Resolver.scala:1274) 
 sangria.execution.Resolver.$anonfun$20(Resolver.scala:813) 
 scala.collection.immutable.Vector1.map(Vector.scala:1872) 
 scala.collection.immutable.Vector1.map(Vector.scala:375) 
 sangria.execution.Resolver.resolveActionsPar(Resolver.scala:1113) 
 sangria.execution.Resolver.resolveFieldsPar(Resolver.scala:49) 
 sangria.execution.Executor.executeOperation(Executor.scala:275) 
 sangria.execution.Executor.$anonfun$5$$anonfun$2$$anonfun$1$$anonfun$1$$anonfun$1(Executor.scala:206) 
 scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470) 
 java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) 
 java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) 
 java.base/java.lang.Thread.run(Thread.java:834)"
```
I am able to successfully get a schema from the running GraphQL server.

Since this particular error relies on having a running GraphQL server, it'll be a little difficult for me to create a miniature isolated code sample.  I might make an attempt at creating an isolated sample, but that would take some time.  If you cannot reproduce, let me know here on the ticket.  (I'll try to monitor the ticket).

---

[...]

PR #1168, opened 2025-01-06
Reproduces future value cast issue for scala 3

mprevel:
Hi,

This PR is related to https://github.com/sangria-graphql/sangria/issues/906
I've added a test to reproduce the encountered issue.
It seems it may be related to derivation since adding the fields explicitly (commented code in the test) works, but I also encounter some strange behaviours while running the tests.

If I use `+testQuick` the new test fails for scala 3 on the 1st run, but it seems to pass on the next one (but I'm not sure it's running with scala 3, logs make me think it runs with scala 2.12)
If I use `+testOnly test.issues.ticket_906.Issue906Spec*`, it fails for scala 3 on each run.

### DIFF
--- a/modules/derivation/src/test/scala/test/issues/ticket_906/Issue906Context.scala
+++ b/modules/derivation/src/test/scala/test/issues/ticket_906/Issue906Context.scala
@@ -1,0 +1,82 @@
+package test.issues.ticket_906
+
+import sangria.macros.derive._
+import sangria.schema._
+
+import scala.concurrent.Future
+
+object Issue906Context {
+
+  @GraphQLName("SampleCaseClass")
+  @GraphQLDescription("A sample case class")
+  final case class SampleCaseClass(
+      name: String
+  ) {
+    @GraphQLField
+    @GraphQLName("myString")
+    def graphqlMyString: String = "Hello World"
+
+    @GraphQLField
+    @GraphQLName("myFutureString")
+    def graphqlMyFutureString: Future[String] = Future.successful("Hello World")
+
+  }
+
+  object SampleCaseClass {
+    val sample: SampleCaseClass = SampleCaseClass("My test")
+  }
+
+  trait MyRepository {
+    def getSampleCaseClass: SampleCaseClass
+  }
+
+  object MyRepository {
+    def sample: MyRepository = new MyRepository {
+      override def getSampleCaseClass: SampleCaseClass = SampleCaseClass.sample
+    }
+  }
+
+  object MySample {
+    val schema: Schema[MyRepository, Unit] = {
+
+      implicit val graphqlSampleCaseClass: ObjectType[MyRepository, SampleCaseClass] =
+        deriveObjectType[MyRepository, SampleCaseClass](
+// it works when adding the fields here but not when using annotations (scala 3)
+//          AddFields(
+//            Field(
+//              "myString",
+//              StringType,
+//              Some("A simple string"),
+//              resolve = _.value.graphqlMyString
+//            ),
+//            Field(
+//              "myFutureString",
+//              StringType,
+//              Some("A simple string"),
+//              resolve = ctx => ctx.value.graphqlMyFutureString
+//            )
+//          )
+        )
+
+      val queries: ObjectType[MyRepository, Unit] = ObjectType(
+        "Query",
+        fields(
+          Field(
+            "mySample",
+            graphqlSampleCaseClass,
+            Some("test"),
+            arguments = Nil,
+            resolve = (ctx: Context[MyRepository, Unit]) => SampleCaseClass.sample
+          )
+        )
+      )
+
+      Schema(
+        query = queries,
+        mutation = None,
+        additionalTypes = Nil
+      )
+    }
+  }
+
+}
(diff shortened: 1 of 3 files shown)


yanns:
Thanks for the reproduction!

The code generated for scala 2:
```scala
  sangria.schema.Field[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass, String, String](
    "myFutureString", 
    implicitly[sangria.macros.derive.GraphQLOutputTypeLookup[String]].graphqlType,
    scala.None,
    scala.collection.immutable.List(),
    ((c: sangria.schema.Context[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass]) => 
      c.value.graphqlMyFutureString),
```

The code generated for scala 3:
```scala
    sangria.schema.Field.apply[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass, scala.Any, scala.Any](
      "myFutureString",
      sangria.macros.derive.GraphQLOutputTypeLookup.outLookup[java.lang.String](sangria.schema.StringType).graphqlType,
      scala.Option.empty[java.lang.String],
      scala.Nil,
      ((`c₃`: sangria.schema.Context[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass]) =>
        ((`value₃`: scala.concurrent.Future[scala.Predef.String]) =>
          sangria.schema.Action.defaultAction[test.issues.ticket_906.Issue906Context.MyRepository, scala.concurrent.Future[scala.Predef.String]](`value₃`)).apply(`c₃`.value.graphqlMyFutureString)),
```

Instead of using the "simple"
```
((c: sangria.schema.Context[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass]) => 
      c.value.graphqlMyFutureString)
```

scala 3 is using something more complex:
```
      ((`c₃`: sangria.schema.Context[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass]) =>
        ((`value₃`: scala.concurrent.Future[scala.Predef.String]) =>
          sangria.schema.Action.defaultAction[test.issues.ticket_906.Issue906Context.MyRepository, scala.concurrent.Future[scala.Predef.String]](`value₃`)).apply(`c₃`.value.graphqlMyFutureString)),
```
It does not call `graphqlMyFutureString` anymore, but encapsulates the body of the function directly in the `resolve`?

yanns:
for context, the PR that implemented the scala macros: https://github.com/sangria-graphql/sangria/pull/900

yanns:
@jchyb pinging you if you have some quick ideas where to look at

robberthcz:
I was looking at the code, and there is method `collectFields` in file `DeriveObjectTypeMacro.scala`. On line 124, the expression `field.onType.memberType(field.method).widen.asType` evaluates probably to `Future[String]` in case of  method `myFutureString`. 

Then on the line 132, the expression `Expr.summon[t => Action[Ctx, t]]` likely evaluates to something like `Expr.summon[Future[String] => Action[Ctx, Future[String]]]`, which finds the implicit method `Action.defaultAction` and not the correct `Action.futureAction`.

It would correspond with the initial ClassCastException, where Sangria resolves the field probably as `Future[Future[String]]` and not the expected `Future[String]` at runtime.

Maybe I'm wrong, I don't really understand macros, so take it with a grain of salt. I will try to debug it more.

yanns:
thanks for the investigation @robberthcz . It seems promising.
I also have no expertise with scala 3 macros, so I guess the only way to make progress is try and error (and debug)

jchyb:
Apologies, it's been some time since I looked into the codebase here, so I cannot pinpoint anything yet. I'll try to see if this is something I can help with. I remember trying to match scala-2's behavior as much as possible, so any deviation here is probably not intended.

jchyb:
I think I see the issue. In scala 2 the generated code does not have to type check, so we can rely on implicit conversions to do some of the job. So while we generate `((c: sangria.schema.Context[test.issues.ticket_906.Issue906Context.MyRepository, test.issues.ticket_906.Issue906Context.SampleCaseClass]) => 
      c.value.graphqlMyFutureString)` there, the `c.value.graphqlMyFutureString` result later gets wrapped in `ReduceAction.futureAction()` by the compiler:
```scala
implicit def futureAction[Ctx, Val](value: Future[Val]): ReduceAction[Ctx, Val] = FutureValue(
    value)
```
In scala 3 the generated code has to type check, so we summon it manually, but for some reason it summons defaultAction:
```scala
  implicit def defaultAction[Ctx, Val](value: Val): ReduceAction[Ctx, Val] = Value(value)
```
So it looks like the summon around line 330 might be wrong. I'll prepare a PR tomorrow

yanns:
@jchyb you don't need to apologize, it's open source, and we're already so thankful to you for having done all the hard work for Scala 3! ❤️ 
And thank you again for your time here.

yanns:
Can you all check that the new sangria release fixes your issue?

robberthcz:
Big thanks. It fixed the issue for us.

mprevel:
> Can you all check that the new sangria release fixes your issue?

It's fixed, thanks.

PR #1169, opened 2025-01-14
Fix issue with wrong Action[Ctx, _] conversion being summoned

### LINKED ISSUE #906: Scala 3: Exception returning Future values from GraphQL Endpoint

jchyb:
The actual change that fixes the issue is very simple, instead of summoning `[fieldType => Action[Ctx, fieldType]]` we summon `[fieldType => Action[Ctx, _]]`, so that we can also summon `ReduceAcrion.futureAction` and `tryAction` (previously we mistakenly limited ourselves to `defaultAction`).

I also allowed myself to simplify the code a bit, wrapping all of the faux-implicit conversions into a separate method, for readability.

Based on the minimization and research done in https://github.com/sangria-graphql/sangria/pull/1168 (I moved the test into the `DeriveObjectTypeMacroSuite`)

Fixes  #906

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -254,6 +254,11 @@ lazy val derivation = project
     name := "sangria-derivation",
     Test / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oF"),
     mimaPreviousArtifacts := Set("org.sangria-graphql" %% "sangria-derivation" % "4.0.0"),
+    mimaBinaryIssueFilters ++= Seq(
+      // internal method
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "sangria.macros.derive.DeriveMacroSupport.unsafeSelectByName")
+    ),
     // Macros
     libraryDependencies ++= (if (isScala3.value) Seq.empty
                              else Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value)),
--- a/modules/derivation/src/main/scala-3/sangria/macros/derive/DeriveMacroSupport.scala
+++ b/modules/derivation/src/main/scala-3/sangria/macros/derive/DeriveMacroSupport.scala
@@ -5,7 +5,7 @@ import sangria.schema.{InputType, OutputType}
 import scala.quoted._
 import scala.reflect.ClassTag
 
-trait DeriveMacroSupport {
+private[derive] trait DeriveMacroSupport {
 
   def reportErrors(using Quotes)(errors: Seq[(PositionPointer, String)]) = {
     import quotes.reflect.report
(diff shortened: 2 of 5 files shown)


yanns (approved):
Thanks a lot! 💯

Issue #1195, opened 2025-05-16
Multi-line Description Can't Contain Nested Lists.


PvtPuddles:
According to the GraphQL spec, the description field can be multi-line and contain markdown-formatted text.

Unfortunately, it appears that Sangria is stripping the whitespace in the description, resulting in nested lists being impossible to create.

The following code:
```scala
val MyArg: Argument[Option[?]] = Argument(
  "myArg",
  OptionInputType(???),
  description = 
    /* language=markdown */
    """My object description
      |
      |- List level one
      |    - List level two
      |""".stripMargin
)
```
results in the following description:

```
"""

My object description

- List level one
- List level two
"""
```

I haven't been able to find a work-around in Markdown to nest the list any way other than leading whitespace.

PvtPuddles:
Our GraphQL API is internal, so I don't consider this urgent (plus, restructuring the description to not require a nested list is doing the job for now).

I am curious of situations where existing users may experience problems with this change. I can think of two that will impact users with poorly-formatted strings (say, not using `.stripMargin`):
- Automated testing may fail since the generated output will change
- If descriptions are not tested, the user will have to manually verify that each description is formatted correctly in the resulting schema (assuming they notice this change in the changelog).

yanns:
I guess that some code might use:
```scala
val MyArg: Argument[Option[?]] = Argument(
  "myArg",
  OptionInputType(???),
  description = 
    /* language=markdown */
    """My object description
      
      - item 1
      - item 2
      """
)
```

and releasing such a change will break all those comments.

PvtPuddles:
> and releasing such a change will break all those comments.

Yes, the resulting markdown would be:

```
"""

My object description

      - Item 1
      - Item 2
"
```

Which would then render something like:

My object description

>  - Item 1
>  - Item 2

yanns:
I've tried to reproduce the issue, but without success for now: https://github.com/sangria-graphql/sangria/pull/1207
Could you check?

PvtPuddles:
Oh, this is annoying.  It appears that Sangria is working exactly as intended.

Running an introspection query manually I get:

```json
"description": "My object description\n\n- List level one\n  - List level two"
```

Which should be rendered correctly.

[...]

PR #1207, opened 2025-06-11
add test about nested lists

yanns:
Try to reproduce https://github.com/sangria-graphql/sangria/issues/1195


yanns:
Merging to test to make sure that we won't break this good behavior.

PR #1217, opened 2025-08-08
Fix O(N^2) memory allocation for parsing newlines

spiliopoulos:
Immutable array appends require an array copy. So memory allocations here grow quadratically with the number of lines in a schema file.

Large repos with a large graphql schemas can have thousands of lines which makes this significant.  

Switch Array -> Mutable ArrayBuffer
Also switches from linear search to binary search

On a 300k GraphQL schema from our production environment that cut parsing time from 16 seconds down to 3 seconds with position tracking and comment parsing.

(Test is not included because I don't have the time to anonymize it, sorry)


spiliopoulos:
@yanns any chance you can take a look?

yanns:
Thanks for the pinging!
I'll have a look later.

PR #1264, opened 2026-03-02
reduce map contention

yanns:
- re-use hash from cacheKeyReversed when possible
- compute the values atomically

### DIFF
--- a/modules/core/src/main/scala/sangria/execution/FieldCollector.scala
+++ b/modules/core/src/main/scala/sangria/execution/FieldCollector.scala
@@ -4,8 +4,8 @@ import sangria.ast.OperationType
 import sangria.ast.SourceMapper
 import sangria.schema._
 import sangria.ast
-import sangria.util.Cache
 
+import java.util.concurrent.ConcurrentHashMap
 import scala.collection.mutable.{ArrayBuffer, Map => MutableMap, Set => MutableSet}
 import scala.util.{Failure, Success, Try}
 
@@ -18,23 +18,30 @@ class FieldCollector[Ctx, Val](
     exceptionHandler: ExceptionHandler) {
 
   private val resultCache =
-    Cache.empty[(ExecutionPath.PathCacheKeyReversed, String), Try[CollectedFields]]
+    new ConcurrentHashMap[
+      ExecutionPath.PathCacheKeyReversed,
+      ConcurrentHashMap[String, Try[CollectedFields]]]
 
   def collectFields(
       path: ExecutionPath,
       tpe: ObjectType[Ctx, _],
       selections: Vector[ast.SelectionContainer]): Try[CollectedFields] =
-    resultCache.getOrElseUpdate(
-      path.cacheKeyReversed -> tpe.name, {
-        val builder: Try[CollectedFieldsBuilder] = Success(new CollectedFieldsBuilder)
+    resultCache
+      .computeIfAbsent(
+        path.cacheKeyReversed,
+        key => new ConcurrentHashMap[String, Try[CollectedFields]])
+      .computeIfAbsent(
+        tpe.name,
+        name => {
+          val builder: Try[CollectedFieldsBuilder] = Success(new CollectedFieldsBuilder)
 
-        selections.foldLeft(builder) { case (acc, s) =>
-          collectFieldsInternal(tpe, s.selections, MutableSet.empty, acc)
+          selections.foldLeft(builder) { case (acc, s) =>
+            collectFieldsInternal(tpe, s.selections, MutableSet.empty, acc)
+          }
+
+          builder.map(_.build)
         }
-
-        builder.map(_.build)
-      }
-    )
+      )
 
   private def collectFieldsInternal(
       tpe: ObjectType[Ctx, _],
--- a/modules/core/src/main/scala/sangria/execution/ValueCollector.scala
+++ b/modules/core/src/main/scala/sangria/execution/ValueCollector.scala
@@ -8,6 +8,7 @@ import sangria.schema._
 import sangria.util.Cache
 import sangria.validation._
 
+import java.util.concurrent.ConcurrentHashMap
 import scala.collection.immutable.VectorBuilder
 import scala.util.{Failure, Success, Try}
 
@@ -24,7 +25,9 @@ class ValueCollector[Ctx, Input](
     new ValueCoercionHelper[Ctx](sourceMapper, deprecationTracker, Some(userContext))
 
   private val argumentCache =
-    Cache.empty[(ExecutionPath.PathCacheKeyReversed, Vector[ast.Argument]), Try[Args]]
+    new ConcurrentHashMap[
+      ExecutionPath.PathCacheKeyReversed,
+      ConcurrentHashMap[Vector[ast.Argument], Try[Args]]]
 
   def getVariableValues(
       definitions: Vector[ast.VariableDefinition],
@@ -106,9 +109,13 @@ class ValueCollector[Ctx, Input](
     if (argumentDefs.isEmpty)
       ValueCollector.emptyArgs
     else
-      argumentCache.getOrElseUpdate(
-        path.cacheKeyReversed -> argumentAsts,
-        getArgumentValues(forAstNode, argumentDefs, argumentAsts, variables))
+      argumentCache
+        .computeIfAbsent(
+          path.cacheKeyReversed,
+          key => new ConcurrentHashMap[Vector[ast.Argument], Try[Args]])
+        .computeIfAbsent(
+          argumentAsts,
+          argumentAsts => getArgumentValues(forAstNode, argumentDefs, argumentAsts, variables))
 
   def getArgumentValues(
       forAstNode: Option[ast.AstNode],


avelx:
[yanns](https://github.com/yanns) - I would assume this should be possible to test this one way or another, or use some kind of benchmarking like: jmeter  or something ?

yanns:
> [yanns](https://github.com/yanns) - I would assume this should be possible to test this one way or another, or use some kind of benchmarking like: jmeter or something ?

Yes, one would need to add some benchmarks. In the scala world, we usually use https://github.com/sbt/sbt-jmh

avelx:
> > [yanns](https://github.com/yanns) - I would assume this should be possible to test this one way or another, or use some kind of benchmarking like: jmeter or something ?
> 
> Yes, one would need to add some benchmarks. In the scala world, we usually use https://github.com/sbt/sbt-jmh

I guess 

> > [yanns](https://github.com/yanns) - I would assume this should be possible to test this one way or another, or use some kind of benchmarking like: jmeter or something ?
> 
> Yes, one would need to add some benchmarks. In the scala world, we usually use https://github.com/sbt/sbt-jmh



> > [yanns](https://github.com/yanns) - I would assume this should be possible to test this one way or another, or use some kind of benchmarking like: jmeter or something ?
> 
> Yes, one would need to add some benchmarks. In the scala world, we usually use https://github.com/sbt/sbt-jmh

sangria.util.Cache is using 

> > [yanns](https://github.com/yanns) - I would assume this should be possible to test this one way or another, or use some kind of benchmarking like: jmeter or something ?
> 
> Yes, one would need to add some benchmarks. In the scala world, we usually use https://github.com/sbt/sbt-jmh

sangria.util.Cache is just a wrapper around ConcurrentHashMap, so not sure if we really need to write tests / benchmarks around this? I would think this is just a draft PR in this case ?

yanns:
> sangria.util.Cache is just a wrapper around ConcurrentHashMap, so not sure if we really need to write tests / benchmarks around this? I would think this is just a draft PR in this case ?

No a benchmark should not check this, but the general impact of the change, with a resolution using synchronous and asynchronous code.

avelx:
> Reference in ne

I understand your intention,
The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with
build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"

There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after

yanns:
> > Reference in ne
> 
> I understand your intention, The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"
> 
> There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after

To be honest, I don't think that we have enough test coverage about a potential impact. That's reason why I haven't continued with this change as I was not feeling confident enough.

avelx:
> > > Reference in ne
> > 
> > 
> > I understand your intention, The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"
> > There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after
> 
> To be honest, I don't think that we have enough test coverage about a potential impact. That's reason why I haven't continued with this change as I was not feeling confident enough.

We need to look for tests around Executor, as I can see there is an "ExecutorSpec" UT,
Do we have Integration Tests (ITs) against sangria-core ? looks like this is something worth to set up to have some level of assurance that these changes works as expected, I will dig around existing UTs, but I don't think this is by nature can cover test part for this change

yanns:
> > > > Reference in ne
> > > 
> > > 
> > > I understand your intention, The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"
> > > There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after
> > 
> > 
> > To be honest, I don't think that we have enough test coverage about a potential impact. That's reason why I haven't continued with this change as I was not feeling confident enough.
> 
> We need to look for tests around Executor, as I can see there is an "ExecutorSpec" UT, Do we have Integration Tests (ITs) against sangria-core ? looks like this is something worth to set up to have some level of assurance that these changes works as expected, I will dig around existing UTs, but I don't think this is by nature can cover test part for this change

One of the change in this PR is how concurrent load of the same key are handled. Before, each load is processed, and each saves a value into the map. With this PR, only one call is doing the loading, saving the value into the map. The other load is waiting for the first one to process.
I'm not sure that tests in `ExecutorSpec` really check how concurrency.

And we don't have any ITs. When I want to asses a new change more deeply, I publish the library locally, and use it in several internal web services. But this process does not scale, and is bind to the fact that I have access to several internal applications using sangria that offers a lot of tests.

avelx:
> > > > > Reference in ne
> > > > 
> > > > 
> > > > I understand your intention, The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"
> > > > There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after
> > > 
> > > 
> > > To be honest, I don't think that we have enough test coverage about a potential impact. That's reason why I haven't continued with this change as I was not feeling confident enough.
> > 
> > 
> > We need to look for tests around Executor, as I can see there is an "ExecutorSpec" UT, Do we have Integration Tests (ITs) against sangria-core ? looks like this is something worth to set up to have some level of assurance that these changes works as expected, I will dig around existing UTs, but I don't think this is by nature can cover test part for this change
> 
> One of the change in this PR is how concurrent load of the same key are handled. Before, each load is processed, and each saves a value into the map. With this PR, only one call is doing the loading, saving the value into the map. The other load is waiting for the first one to process. I'm not sure that tests in `ExecutorSpec` really check how concurrency.
> 
> And we don't have any ITs. When I want to asses a new change more deeply, I publish the library locally, and use it in several internal web services. But this process does not scale, and is bind to the fact that I have access to several internal applications using sangria that offers a lot of tests.

I think we can just raise an Github issue and park this conversation as there are quite a list of challanges here like to add ITs to the coRe project etc. Let me know if you happy for me to do this.

yanns:
> > > > > > Reference in ne
> > > > > 
> > > > > 
> > > > > I understand your intention, The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"
> > > > > There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after
> > > > 
> > > > 
> > > > To be honest, I don't think that we have enough test coverage about a potential impact. That's reason why I haven't continued with this change as I was not feeling confident enough.
> > > 
> > > 
> > > We need to look for tests around Executor, as I can see there is an "ExecutorSpec" UT, Do we have Integration Tests (ITs) against sangria-core ? looks like this is something worth to set up to have some level of assurance that these changes works as expected, I will dig around existing UTs, but I don't think this is by nature can cover test part for this change
> > 
> > 
> > One of the change in this PR is how concurrent load of the same key are handled. Before, each load is processed, and each saves a value into the map. With this PR, only one call is doing the loading, saving the value into the map. The other load is waiting for the first one to process. I'm not sure that tests in `ExecutorSpec` really check how concurrency.
> > And we don't have any ITs. When I want to asses a new change more deeply, I publish the library locally, and use it in several internal web services. But this process does not scale, and is bind to the fact that I have access to several internal applications using sangria that offers a lot of tests.
> 
> I think we can just raise an Github issue and park this conversation as there are quite a list of challanges here like to add ITs to the coRe project etc. Let me know if you happy for me to do this.

Thanks for proposing. Any help appreciated here!

avelx:
> > > > > > > Reference in ne
> > > > > > 
> > > > > > 
> > > > > > I understand your intention, The only difference I see is that we are replacing custom-made _getElseOrUpdate_ with build in "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)"
> > > > > > There must be existing tests we can use to check if there is no impact, or we need to add a new one to assess impact and see how **_Executor_** behaves after
> > > > > 
> > > > > 
> > > > > To be honest, I don't think that we have enough test coverage about a potential impact. That's reason why I haven't continued with this change as I was not feeling confident enough.
> > > > 
> > > > 
> > > > We need to look for tests around Executor, as I can see there is an "ExecutorSpec" UT, Do we have Integration Tests (ITs) against sangria-core ? looks like this is something worth to set up to have some level of assurance that these changes works as expected, I will dig around existing UTs, but I don't think this is by nature can cover test part for this change
> > > 
> > > 
> > > One of the change in this PR is how concurrent load of the same key are handled. Before, each load is processed, and each saves a value into the map. With this PR, only one call is doing the loading, saving the value into the map. The other load is waiting for the first one to process. I'm not sure that tests in `ExecutorSpec` really check how concurrency.
> > > And we don't have any ITs. When I want to asses a new change more deeply, I publish the library locally, and use it in several internal web services. But this process does not scale, and is bind to the fact that I have access to several internal applications using sangria that offers a lot of tests.
> > 
> > 
> > I think we can just raise an Github issue and park this conversation as there are quite a list of challanges here like to add ITs to the coRe project etc. Let me know if you happy for me to do this.
> 
> Thanks for proposing. Any help appreciated here!

Here is an issue for this work: https://github.com/sangria-graphql/sangria/issues/1284

Feel free to comment on the issue itself, I think we can close this thread.

PR #1267, opened 2026-03-06
Fix Float rejecting decimals like 144.75999999999999

TidyMaze:
Float inputs like `144.75999999999999` were being rejected with "too big to fit in double" even though they fit in a double. The old logic required an exact decimal round-trip (`isDecimalDouble`), so any value that doesn’t round-trip exactly (common with JSON or GraphQL literals) failed.

We now coerce to the nearest double and only reject when the value actually overflows to infinity. So those numbers work as expected, and the behaviour matches what people expect from a Float type. Fully backward compatible.

Technical change: `FloatType` and `ResolverBasedAstSchemaBuilder` use `isFinite(d.doubleValue)` instead of `isDecimalDouble` for `BigDecimal`. Tests added in CoercionSpec, ValueCoercionHelperSpec, and VariablesSpec.


yanns:
Can you reformat with `scalafmtAll`? Thank yoy

TidyMaze:
we implemented a workaround in our codebase, closing that

PR #1285, opened 2026-04-16
WIP: Add integration tests

avelx:


it/test/controllers/TestExecutorController.scala
```diff
@@ -0,0 +1,31 @@
...
+                hero {
+                  name
+                }
+              }
+              """
+
+    val result: Future[Any] =  Executor
+        .execute(StarWarsSchema, query, new CharacterRepo, deferredResolver = new FriendsResolver)
+
+    result.map { x =>
```
yanns: We could return some json instead to make it more realistic.

it/test/controllers/TestExecutorController.scala
```diff
@@ -0,0 +1,31 @@
...
+      Ok(s"ok:$x")
+    }
+  }
+}
```
yanns: Nitpick: configure your text editor to always add a new line.

it/test/query/StarWarsQueryISpec.scala
```diff
@@ -0,0 +1,34 @@
...
+import play.api.mvc.{Result, Results}
+import play.api.test.Helpers.{contentAsString, defaultAwaitTimeout}
+import play.api.test.{FakeRequest, Helpers}
+
+import scala.concurrent.Future
+
+/*
+  Similar to the tests we have within => StarWarsQuerySpec but via controller
+ */
+class StarWarsQuerySpecISpec extends PlaySpec with Results {
```
yanns: To avoid relying on `PlaySpec` (one more dependency to take care of), could we "just" write assertions on a function returning a `Future[Json]`?

yanns (commented):
the direction is good!

it/test/controllers/TestExecutorController.scala
avelx: change applied

it/test/controllers/TestExecutorController.scala
avelx: next scalaFmt command applied:
>sbt it/scalafmtAll

it/test/query/StarWarsQueryISpec.scala
avelx: As I understand, you are ok with using Play as an environment, but just drop PlaySpec specifically? 
yeh, we I can do this

avelx: suggested approach applied;
Don't see much difference, but let it be this way

yanns: I'd actually prefer not using Play as an environment to avoid adding more dependencies

avelx: What surrounding environment would you suggest? as I understand its not worth to continue experimenting with the Test play controller ?

yanns: We can emulate an environment with a main function returning a Future[Json].

avelx: ok, I've dropped Play dependency and use Future[String] to have more flexibility ;

so, now  the only difference between StarWarsQuerySpecISpec and StarWarsQuerySpec is that we wrapped Executed within TestEnvironment

PR #1302, opened 2026-07-03
Fix Scala 3 path-dependent method return derivation

### LINKED ISSUE #1298: Scala 3 deriveObjectType fails for methods returning path-dependent F[A]
## Summary

`deriveObjectType` / `deriveContextObjectType` on Scala 3 can fail when deriving a type whose annotated methods return an effect type through a path-dependent type parameter, for example an inner `Queries[F]`/`Fields[F]` holder derived from an outer `Resource[F]`.

The practical shape is:

```scala
import cats.effect.Async
import sangria.macros.derive.*
import sangria.schema.*

class Resource[F[_]: Async]:
  class Queries[G[_]: Async]:
    @GraphQLField
    def value: G[Int] = Async[G].pure(42)

  val queryType: ObjectType[Unit, Queries[F]] =
    deriveObjectType[Unit, Queries[F]]()
```

With real-world code this can produce an `ExprCastException` because the generated resolver expression has a type like `Resource.this.F[A]`, while the macro casts it using the method raw declared return type like `Queries.this.G[A]` / `Queries.this.F[A]`.

## Why this happens

In the Scala 3 derivation macro, field schema typing already uses an as-seen-from type:

```scala
val fieldType = methodResultType(field.onType.memberType(field.method).widen)
```

But the resolver branch later uses the raw `DefDef.returnTpt.tpe` when calling `asExprOf[t]`:

```scala
case d: DefDef =>
  d.returnTpt.tpe.asType match
    case '[t] =>
      wrappedInActionConversion[Ctx, t](
        Apply(...).asExprOf[t]
      )
```

For path-dependent higher-kinded effect parameters these two types are equivalent for users, but not identical enough for the quoted cast.

## Impact

This blocks deriving GraphQL field holders written in a tagless-final style, especially code organized as reusable `Queries[F]`, `Mutations[F]`, or `Fields[F]` classes. We have this pattern in production GraphQL APIs, and the workaround is to hand-write or replace derived object types, which can easily drop resolver fields from the schema.

## Proposed fix

Use the same as-seen-from method result type for the generated resolver cast that is already used for field type discovery, instead of the raw `DefDef.returnTpt.tpe`.

A good regression test would cover an outer `F[_]`, an inner field holder parameterized by `F[_]`, an `@GraphQLField` method returning `F[Int]`, and successful derivation/execution on Scala 3.

eugkhp:
## Summary

Fixes #1298.

This changes the Scala 3 object derivation macro to use the method return type as seen from the derived owner type when casting generated resolver expressions. The field type path already used this contextualized type; the resolver path still used the raw `DefDef.returnTpt.tpe`, which can differ for path-dependent higher-kinded return types.

## Why

For field holders such as `Queries[F]` derived from an outer resource with a path-dependent `F`, the generated resolver expression may type as the outer `F[A]` while the raw method declaration type is the inner type parameter `G[A]`. Using the owner-contextualized member type keeps both sides aligned.

## Validation

- `JAVA_HOME=$(cs java-home --jvm temurin:17) sbt '++3.3.7' 'sangria-derivation/Test/testOnly sangria.macros.derive.DeriveObjectTypeMacroSpec'`
- `JAVA_HOME=$(cs java-home --jvm temurin:17) sbt scalafmtCheckAll`

### DIFF
--- a/modules/derivation/src/main/scala-3/sangria/macros/derive/DeriveObjectTypeMacro.scala
+++ b/modules/derivation/src/main/scala-3/sangria/macros/derive/DeriveObjectTypeMacro.scala
@@ -143,11 +143,6 @@ class DeriveObjectTypeMacro(using globalQuotes: Quotes) extends DeriveMacroSuppo
             else
               fieldWithArguments(config, field, ctxType, valType, ctxValType)
 
-          @tailrec
-          def methodResultType(typeRepr: TypeRepr): TypeRepr = typeRepr match
-            case MethodType(_, _, retType) => methodResultType(retType)
-            case retType => retType
-
           // Contextualize the method type with respect to the enclosing type
           val fieldType = methodResultType(field.onType.memberType(field.method).widen)
           val actualFieldType = findActualFieldType(fieldType)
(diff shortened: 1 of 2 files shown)


yanns:
Thanks for your contribution!
Can you check the formatting? The CI is failing because of that: https://github.com/sangria-graphql/sangria/actions/runs/28673710796/job/85252067418?pr=1302

eugkhp:
Thank you! Fixed!

yanns (approved):
👍

Issue #1299, opened 2026-07-03
Add cats-effect derivation support for fields returning F[A]


eugkhp:
## Summary

`sangria-cats-effect-experimental` currently provides an `AsyncExecutionScheme`, `AsyncValue`, and resolver integration, so an execution can produce `F[Json]` and explicit fields can return `AsyncValue`. What is still missing is ergonomic derivation support for annotated methods returning `IO[A]` or generic `F[A]`.

In production code, the common shape is:

```scala
import cats.effect.IO
import sangria.macros.derive.*
import sangria.schema.*

case class User(id: Int, name: String)

class Queries:
  @GraphQLField
  def users: IO[List[User]] = IO.pure(List(User(1, "Ada")))

val queryType: ObjectType[Unit, Queries] =
  deriveObjectType[Unit, Queries]()
```

Expected behavior: derivation should be able to infer the GraphQL output type from `A`, while the cats-effect integration should resolve the `F[A]` effect at execution time. This should work for common Sangria containers and outputs, including:

- `F[List[A]]` / `F[Seq[A]]` / `F[Vector[A]]`
- `F[Option[A]]`
- `F[A]` for object types
- `F[Int]`, `F[String]`, etc. for scalar types
- `F[Unit]` for mutation-like fields

## Current workaround

Users have to manually define fields or carry a local support layer that adds `GraphQLOutputTypeLookup[F[A]]`, derives object types through wrapper methods, and converts returned effects into Sangria actions.

## Proposed direction

Add a small cats-effect derivation ergonomics layer to `sangria-cats-effect-experimental`:

- a `GraphQLOutputTypeLookup[F[A]]` that delegates to the lookup for `A`
- an action conversion from `F[A]` to the cats-effect action type used by the experimental resolver
- regression tests showing derived fields returning `IO[List[A]]`, `IO[Option[A]]`, `IO[A]`, scalar `IO` values, and `IO[Unit]` execute successfully

This would make the experimental module useful for tagless-final GraphQL APIs without each application carrying its own Sangria derivation bridge.

yanns:
@eugkhp is this issue solved by https://github.com/sangria-graphql/sangria/pull/1303 or do you think there's still some work to do?

eugkhp:
Yeah, it does!

eugkhp:
Closed by #1303.

PR #1303, opened 2026-07-03
Add Scala 3 cats-effect derivation lookups

eugkhp:
## Summary

Part of #1299.

This adds Scala 3 derivation support for cats-effect fields by exposing `GraphQLOutputTypeLookup` instances from `AsyncValue._`:

- `GraphQLOutputTypeLookup[IO[A]]` delegates to `GraphQLOutputTypeLookup[A]`
- generic `GraphQLOutputTypeLookup[F[A]]` delegates to `A` when an `Async[F]` is available

The cats-effect experimental module now depends on `sangria-derivation` so these helpers can live with the existing `AsyncValue` action conversion import.

## Scope

The regression test is Scala 3-only because Scala 3 derivation explicitly summons the field action conversion and can therefore resolve `IO[A]` through `AsyncValue`. Scala 2 derivation uses the method return type as the `Field` result type for argless methods, so a lookup alone is not sufficient to unwrap `IO[A]` there without a separate macro change.

## Validation

- `JAVA_HOME=$(cs java-home --jvm temurin:17) sbt '++3.3.7' 'sangria-cats-effect-experimental/Test/testOnly sangria.catseffect.IOExecutionSchemeSpec sangria.catseffect.IOExecutionSchemeDerivationSpec'`
- `JAVA_HOME=$(cs java-home --jvm temurin:17) sbt '++2.13.18' 'sangria-cats-effect-experimental/Test/testOnly sangria.catseffect.IOExecutionSchemeSpec'`
- `JAVA_HOME=$(cs java-home --jvm temurin:17) sbt scalafmtCheckAll`

### DIFF
--- a/modules/cats-effect-experimental/src/main/scala/sangria/catseffect/schema/AsyncValue.scala
+++ b/modules/cats-effect-experimental/src/main/scala/sangria/catseffect/schema/AsyncValue.scala
@@ -1,7 +1,8 @@
 package sangria.catseffect.schema
 
-import cats.effect.Async
-import sangria.schema.LeafAction
+import cats.effect.{Async, IO}
+import sangria.macros.derive.GraphQLOutputTypeLookup
+import sangria.schema.{LeafAction, OutputType}
 
 import scala.concurrent.ExecutionContext
 import scala.language.implicitConversions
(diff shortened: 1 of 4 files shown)


modules/cats-effect-experimental/src/main/scala/sangria/catseffect/schema/AsyncValue.scala
```diff
@@ -12,7 +13,23 @@ case class AsyncValue[Ctx, Val, F[_]: Async](value: F[Val]) extends LeafAction[C
...
+object AsyncValue extends AsyncValueLowPriority {
   implicit def asyncAction[Ctx, Val, F[_]: Async](value: F[Val]): LeafAction[Ctx, Val] =
     AsyncValue(value)
+
+  implicit def ioOutputTypeLookup[A](implicit
+      inner: GraphQLOutputTypeLookup[A]): GraphQLOutputTypeLookup[IO[A]] =
+    new GraphQLOutputTypeLookup[IO[A]] {
+      override val graphqlType: OutputType[IO[A]] =
+        inner.graphqlType.asInstanceOf[OutputType[IO[A]]]
+    }
```
yanns: If I remove this, the code is still compiling and tests are passing. Why should we need it? `asyncOutputTypeLookup` might be enough?

eugkhp: You’re right, thanks. I added it because our local AvantStay support layer had both the IO-specific and generic F[_]: Async lookups, but after re-testing, the generic one is enough. Removed it in 9c5bce51.

For context, we use Sangria pretty extensively at AvantStay with a bunch of cats-effect / tagless-final glue around derivation and execution. Happy to share or contribute more of it, though some parts are more opinionated/app-specific than this small lookup.

yanns: > For context, we use Sangria pretty extensively at AvantStay with a bunch of cats-effect / tagless-final glue around derivation and execution. Happy to share or contribute more of it, though some parts are more opinionated/app-specific than this small lookup.

Good to know. We're also using Sangria with cats-effets but without tagless-final.
Happy to take any contribution, especially when they're as good as yours, with tests included 😉

yanns (approved):

